PR #21, opened 2026-08-01
feat: expected-rate alerting: WARN lines + pulse-check CLI (ROADMAP R1)

TanayK07:
## Summary

R1 turns the logger into a monitor: declare what "healthy" means, get `WARN` lines when reality disagrees, and gate watchdogs/CI on a single exit code, zero per-node code, zero hot-path cost.

**Do not merge yet, for review.**

## What ships

- **Spec file** via `ROS_TOPIC_STATS_EXPECTED`, a documented YAML subset (flow-map topic rules, flow/block node lists, comments) parsed by ~200 lines of dependency-free C++, so no YAML library enters an LD_PRELOAD shim:
  ```yaml
  topics:
    /scan:   {min_hz: 18, max_hz: 22, side: recv}
    /points: {min_hz: 25, transport: intra}
  nodes: [/perception, /planner]
  ```
- **WARN lines**, additive to the window block (existing parsers unaffected):
  ```
  WARN TOPIC /scan hz=1.200000 expected=[18,22]
  WARN NODE /planner missing
  ```
  Evaluated at **flush time only**; first non-empty window grace-skipped (attach ramp-up would cry wolf); each process judges only endpoints it hosts; malformed/unreadable specs warn once on stderr and disable alerting, never crash the host.
- **`pulse-check` CLI** (no ROS dependency, links the same core): `pulse-check --spec expected.yaml /tmp/pulse.*.log` → exit 0 pass / 1 violations / 2 bad input. Re-derives the verdict from raw rates of each log's LAST window; offline it owns the whole picture, so a spec topic in no log = violation at 0 Hz. Installed to `lib/ros2_pulse/`.

## Design notes

- `rate_spec` (parser + evaluator) and `log_reader` (inverts `formatWindow`) live in the pure core: unit-tested in isolation, shared verbatim by probe and CLI (DRY).
- `TopicRegistry::knownNodes()` distinguishes "aged out → warn" from "never ours → skip".
- `formatWindow` gains a defaulted `warnings` param, empty list keeps output byte-identical to pre-R1.

## TDD

- Red commit: 12 failing tests (parser grammar + hard errors with line numbers, pinned WARN bytes, side/transport selection, probe-vs-offline missing semantics, node liveness, log round-trip, WARN placement).
- Green commit: all pass; then CLI + integration.

## Verification

- Unit: 75/75 × fast / TSan / ASan-UBSan standalone lanes
- colcon on humble: **104 tests, 0 failures** (1 expected distro skip), includes 4 new integration tests: WARN appearance + first-window grace, satisfied-spec silence, malformed-spec resilience, all four pulse-check verdicts
- Docs: README section + env row, ROADMAP R1 crossed off; also fixed the stale "16-slot / ~3ns" how-it-works text left over from #19


src/probe/interposers.cpp
```diff
@@ -63,6 +67,40 @@ auto resolveOutputPath() -> std::string {
...
+        std::fprintf(stderr,
+                     "[ros2_pulse] cannot read ROS_TOPIC_STATS_EXPECTED '%s' (%s) — "
+                     "expected-rate alerting disabled\n",
+                     path, std::strerror(errno));
+        return std::nullopt;
+    }
+    std::string text;
+    char buf[4096];
+    size_t n = 0;
+    while ((n = std::fread(buf, 1, sizeof(buf), f)) > 0) {
```
TanayK07, line 89: ## 🔬 ISSUE: unbounded spec-file read inside a tracepoint-reached constructor

**Severity:** MEDIUM — a misconfigured `ROS_TOPIC_STATS_EXPECTED` path can OOM or `std::terminate` every LD_PRELOADed process on the robot at startup.

### The Problem

```cpp
std::string text;
char buf[4096];
size_t n = 0;
while ((n = std::fread(buf, 1, sizeof(buf), f)) > 0) {
    text.append(buf, n);
}
```

`loadRateSpec()` runs in the `ProbeRuntime` constructor, which is reached from an `extern "C"` tracepoint called by rclcpp (same context as `parsePeriodSeconds`, whose header documents the no-throw contract from KNOWN_ISSUES #5). The read loop has no size cap: it slurps whatever the env var points at, in full, into a `std::string`.

### Theory / Proof

Two failure modes, both fleet-wide because `LD_PRELOAD` + env vars are typically set for *every* process in a launch:

1. `std::string::append` throws `std::bad_alloc` when allocation fails. The exception unwinds out of the singleton constructor, across the `extern "C"` interposer boundary into rclcpp code not compiled to expect it → `std::terminate` of the **host** process. This is exactly the failure class issue #5 eliminated for `ROS_TOPIC_STATISTICS_PUBLISH_PERIOD` (`env_config.hpp` documents the rationale).
2. Even when allocation succeeds, pointing the var at a large file (a rosbag, a core dump, `/dev/urandom` never returns EOF at all) burns RAM/time in every process simultaneously at startup.

Today: wrong-but-missing paths are handled (`fopen` fails → warn-once + disable ✅). What's not handled is wrong-but-*existing* paths: `ROS_TOPIC_STATS_EXPECTED=/data/run1.bag` is one tab-completion away from `/data/run1.yaml`.

### Research

Defensive size caps on operator-supplied config are standard in preload/agent tooling; e.g. glibc itself hard-limits `LD_PRELOAD` list parsing, and observability agents (Datadog's `dd-trace`, OTel SDKs) cap config file reads. A realistic spec (hundreds of topics) is a few tens of KB.

### Recommendation

Cap the read (1 MiB is generous) and warn+disable past it — same degrade-never-crash stance as the rest of the loader; a fixed `fread` into a capped buffer also removes the throwing `append` from the growth path:

```cpp
constexpr size_t kMaxSpecBytes = 1 << 20;  // 1 MiB — a real spec is a few KB
while ((n = std::fread(buf, 1, sizeof(buf), f)) > 0) {
    if (text.size() + n > kMaxSpecBytes) {
        std::fclose(f);
        std::fprintf(stderr, "[ros2_pulse] ROS_TOPIC_STATS_EXPECTED '%s' exceeds %zu bytes — "
                     "not a spec? expected-rate alerting disabled\n", path, kMaxSpecBytes);
        return std::nullopt;
    }
    text.append(buf, n);
}
```

Bonus operator UX while touching this: the `[ros2_pulse] active` banner in `ensureStarted()` doesn't say whether a spec armed — printing `spec: N topics, M nodes` there makes "why no WARNs?" debuggable at a glance.

src/probe/interposers.cpp
```diff
@@ -216,8 +255,16 @@ class ProbeRuntime {
...
         // and the TOPIC emit decision (incl. the idle-topic policy gated by ROS_PULSE_EMIT_IDLE)
         // lives in the pure core so it stays unit-testable (issue #7).
-        const std::string block =
-            formatWindow(stats, nodes, static_cast<long long>(ns), window_s, m_emit_idle);
+        // Expected-rate alerting (ROADMAP R1): evaluated here at flush time only — the hot path
+        // never sees the spec. The FIRST non-empty window is grace-skipped: the probe attaches
+        // mid-flight, so that window's rates are ramp-up partials that would cry wolf on start.
+        const uint64_t window_index = m_windows_flushed.fetch_add(1, std::memory_order_relaxed);
+        std::vector<std::string> warnings;
+        if (m_spec.has_value() && window_index > 0) {
```
TanayK07, line 263: ## 🔬 ISSUE: exit-time partial window can emit spurious WARNs (probe and pulse-check)

**Severity:** MEDIUM — alert fatigue: every graceful shutdown can log violations that aren't, and pulse-check judges exactly that window.

### The Problem

```cpp
const uint64_t window_index = m_windows_flushed.fetch_add(1, std::memory_order_relaxed);
std::vector<std::string> warnings;
if (m_spec.has_value() && window_index > 0) {
    warnings = evaluateRateSpec(*m_spec, stats, nodes, m_registry.knownNodes());
}
```

The grace guard handles the *attach* transient (first window) but not the *detach* transient: `shutdownAtExit()` calls this same `flush()` for the final partial window. If the node stopped publishing before process exit — `rclcpp::shutdown()` cancels timers, executor drains, destructors run, and only then does `atexit` fire — the tail window contains a silent gap. The measured-window division (issue #8b) keeps rates honest *for the time traffic flowed*, but the gap still drags `count / window_s` below `min_hz`.

### Theory / Proof

Concrete: 20 Hz topic, 5 s period, operator Ctrl-C at t=2.0 into the window; rclcpp teardown takes 1.5 s; atexit flush at t=3.5. Count = 40, window_s = 3.5 → 11.4 Hz < min 18 → `WARN TOPIC /scan hz=11.43 expected=[18,22]` on a perfectly healthy shutdown. Compounding: `pulse-check` deliberately evaluates **the LAST window of each log** (tools/pulse_check.cpp header), which after any clean stop is exactly this partial — so a CI job that runs a stack, stops it, then gates on pulse-check can flake red.

The integration tests pass because `run_single` SIGTERMs while the talker is still publishing full-rate — the tail window has no silent gap. Any user whose nodes drain before exit (lifecycle nodes, launch shutdown ordering) will see WARNs at every stop.

### Research

Monitoring systems treat shutdown as a distinct state precisely to avoid this: systemd suppresses watchdog checks once `Stopping=` is entered; Prometheus alerting uses `for:` durations so single-scrape dips don't page. `diagnostic_updater`'s FrequencyStatus has the same partial-window weakness, which is one reason its windows are short and its consumers debounce.

### Recommendation

The probe knows which flush is the exit flush — pass that through and skip evaluation (rates still get logged; only the alert judgement is suppressed):

```cpp
void shutdownAtExit() {
    if (m_timer) { m_timer->stop(); }
    flush(/*exiting=*/true);
}
void flush(bool exiting = false) {
    ...
    if (m_spec.has_value() && window_index > 0 && !exiting) {
```

For pulse-check, either document "judge only steady-state logs" or add `--skip-last` for post-run CI use. A `for: N consecutive windows` debounce (only warn after K consecutive violating windows) would solve both transients generally — reasonable R1.1 follow-up, not needed for this PR.

src/core/rate_spec.cpp
```diff
@@ -0,0 +1,288 @@
...
+        }
+    } else {
+        return "unknown key '" + key + "' (expected min_hz, max_hz, side, transport)";
+    }
+    return {};
+}
+
+// The rate field a rule constrains. kAny sums both transports: "how fast is this topic,
+// however it travels" — the split is a transport detail the operator usually doesn't spec.
+auto observedHz(const sRateRule& rule, const sTopicStat& s) -> double {
```
TanayK07, line 110: ## 🔬 GAP ANALYSIS: `transport: any` sums buckets that can double-count one produce rate

**Severity:** MEDIUM — the default transport on the publish side can report 2× the real rate on Jazzy+, silently inverting `max_hz` semantics.

### The Problem

```cpp
switch (rule.transport) {
    case eRateTransport::kInter: return inter;
    case eRateTransport::kIntra: return intra;
    default: return inter + intra;   // kAny
}
```

For `side: recv`, summing is right: each delivery lands in exactly one bucket (`callback_start` fires once per callback, `is_intra_process` selects the bucket), so inter + intra = total deliveries.

For `side: pub` on Jazzy+, the buckets are **not disjoint**: one `publish()` call on a publisher with both in-process and out-of-process subscribers fires `rclcpp_intra_publish` (IPC path) *and* `rcl_publish` (RMW path for the remote subscribers). rclcpp's `PublisherBase`/`IntraProcessManager` split delivery: intra goes through the IPM, everything else still goes down through rcl — both tracepoints fire for the same logical message.

### Theory / Proof

50 Hz camera driver, IPC on, one composed consumer in-process plus one external recorder: `pub_intra_hz = 50`, `pub_inter_hz = 50`. Rule `/image: {min_hz: 45, max_hz: 55, side: pub}` (transport defaults to `any`) → observed 100 Hz → `WARN TOPIC /image hz=100.0 expected=[45,55]` on a healthy 50 Hz stream. The same shape recurs in pulse-check's cross-log merge (`mergeStat` sums per field): two subscriber *processes* on one topic honestly sum to 2× on the recv side — correct as "total deliveries", surprising as "topic rate" the moment someone sets `max_hz`.

Today nothing breaks: Humble has no intra-publish tracepoint (pub_intra always 0) and the tests only exercise single-transport topics. The first Jazzy user with IPC + mixed subscribers and a `max_hz` bound gets a false alarm that looks like a message storm.

### Research

`ros2 topic hz` and rclcpp's TopicStatistics both measure at a *single* subscription, so "rate" is per-observer and unambiguous. The moment a monitor aggregates across observers/transports it has to pick a semantic — Prometheus solves this by making the aggregation explicit in the query (`sum without(transport)` vs `max`). An implicit lossy sum as the *default* is the trap.

### Recommendation

For `side: pub`, make `kAny` mean "the produce rate" rather than "the bucket sum" — `max(inter, intra)` matches the physical event count whenever the same message rides both paths, and equals the single bucket when only one path is active:

```cpp
if (rule.side == eRateSide::kPub && rule.transport == eRateTransport::kAny) {
    return inter > intra ? inter : intra;   // one publish() may fire both tracepoints
}
```

(recv keeps the sum — deliveries are disjoint there). Whatever the decision, pin it in a unit test with both pub buckets non-zero and document it in the README's transport line; today that case has no test.

src/core/rate_spec.cpp
```diff
@@ -0,0 +1,288 @@
...
+                if (colon == std::string::npos || brace == std::string::npos ||
+                    line.back() != '}' || brace < colon) {
+                    return fail(error, line_no,
+                                "expected '<topic>: {min_hz: ..., ...}' inside 'topics:'");
+                }
+                const std::string name = trim(line.substr(0, colon));
+                if (name.empty()) {
+                    return fail(error, line_no, "empty topic name");
+                }
+                if (!seen_topics.insert(name).second) {
```
TanayK07, line 220: ## 🔬 GAP ANALYSIS: one rule per topic — pub and recv cannot both be constrained

**Severity:** LOW — expressible-spec gap that surfaces as a confusing hard error.

### The Problem

```cpp
if (!seen_topics.insert(name).second) {
    return fail(error, line_no, "duplicate topic '" + name + "'");
}
```

The dedup is keyed on topic name alone, so the natural spec for "the driver must publish ~20 Hz AND my process must receive ~20 Hz" is unwritable:

```yaml
topics:
  /scan: {min_hz: 18, side: pub}
  /scan: {min_hz: 18, side: recv}   # line 3: duplicate topic '/scan'
```

### Theory / Proof

`sRateSpec::topics` is already `vector<pair<name, rule>>` — nothing downstream requires uniqueness; `evaluateRateSpec` iterates the vector and would handle two `/scan` rules with different sides correctly as written. The uniqueness constraint exists only in the parser. Monitoring both sides of one topic in one process is a first-class use case for this tool — it is literally the same-process pub+sub scenario that KNOWN_ISSUES #1 split the counter buckets for.

Today a duplicate is *usually* a copy-paste mistake, and the hard error is right for the same-side case. The two-sided case fails with a message that suggests the spec is wrong rather than the grammar being too small.

### Research

diagnostic_updater composes multiple `FrequencyStatus` tasks on one topic without conflict; Prometheus alert rules on one series are unlimited. Uniqueness-per-name is the outlier.

### Recommendation

Key the dedup on `(name, side, transport)` so identical rules still error but complementary ones pass:

```cpp
std::unordered_set<std::string> seen_rules;
...
const std::string rule_key = name + '\x01' + std::to_string(static_cast<int>(rule.side)) +
                             '\x01' + std::to_string(static_cast<int>(rule.transport));
if (!seen_rules.insert(rule_key).second) {
    return fail(error, line_no, "duplicate rule for '" + name + "' (same side+transport)");
}
```

Note the check must move *after* the rule body is parsed. One parser test (two sides on one topic parse; two identical entries still fail) locks it in.

src/core/rate_spec.cpp
```diff
@@ -0,0 +1,288 @@
...
+        if (x == s) {
+            return true;
+        }
+    }
+    return false;
+}
+
+}  // namespace
+
+auto parseRateSpec(const std::string& text, std::string& error) -> std::optional<sRateSpec> {
```
TanayK07, line 145: ## 🔬 ISSUE: UTF-8 BOM at the top of a spec disables alerting with a misleading error

**Severity:** LOW — Windows-authored specs silently degrade to "no alerting" with an error that blames the wrong thing.

### The Problem

```cpp
auto parseRateSpec(const std::string& text, std::string& error) -> std::optional<sRateSpec> {
```

The line lexer classifies a line as "indented" iff `raw[0]` is space/tab, and matches top-level keys byte-exactly (`line == "topics:"`). A file that begins with the UTF-8 BOM (`EF BB BF`) — the default for Windows Notepad and a common VS Code/PowerShell artifact — turns the first line into `\xEF\xBB\xBFtopics:`, which is neither indented nor equal to `topics:`, so parsing fails with:

```
line 1: unknown top-level entry 'topics:'   <- looks identical to the correct spelling
```

`trim()` only strips ` \t\r`, so the BOM survives into both the comparison and the error message (where it is invisible).

### Theory / Proof

Unicode standard §23.8: U+FEFF at stream start is a signature, not content; consumers of UTF-8 text "should" tolerate and strip it. The operator-facing symptom is nasty precisely because the error text *prints* the invisible bytes: the message shows `'topics:'` — indistinguishable from the working spelling — and the probe then runs with alerting off (warn-once on stderr, easily missed in a launch's output).

Linux-authored specs are fine. The first spec written on a Windows workstation and scp'd to the robot loses alerting with no actionable diagnostic — the failure mode this feature exists to prevent (silent absence of monitoring).

### Research

Every YAML library strips the BOM (libyaml `yaml_parser_determine_encoding`, SnakeYAML, PyYAML); Go's `encoding/json` and .NET text readers do the same. A subset parser that replaces a YAML lib inherits this obligation.

### Recommendation

One strip at entry:

```cpp
auto parseRateSpec(const std::string& text_in, std::string& error) -> std::optional<sRateSpec> {
    std::string_view text{text_in};
    if (text.size() >= 3 && text.substr(0, 3) == "\xEF\xBB\xBF") {
        text.remove_prefix(3);  // UTF-8 BOM: signature, not content (Unicode 23.8)
    }
```

plus a two-line parser test (`"\xEF\xBB\xBFtopics:\n  /x: {min_hz: 1}\n"` parses). CRLF is already covered by `trim()` — worth a test asserting that too, so the Windows path stays green as the parser evolves.

src/probe/interposers.cpp
TanayK07: The resource bound is worth fixing, but not for the reason given — and the fix needs to be wider.

Reproduced the loop standalone: `/dev/urandom` and `/dev/zero` never return 0 from `fread`, so it genuinely never terminates, and a 2 GiB input costs 2.76 s and 2.05 GiB RSS *per process* inside `rcl_node_init`. Real problem.

The `std::terminate` framing doesn't hold up though. `extern "C"` governs linkage, not unwinding — an exception propagates fine through rcl's C frames on the Itanium ABI; death would come from no handler up the stack, not a boundary incompatibility. More to the point, the same constructor already allocates at `new ProbeRuntime()` (:116), `m_out_path` (:223), `new Timer`/`std::thread` (:137-139) and throughout `parseRateSpec`, so a byte cap buys no no-throw guarantee. This isn't issue #5's class — that was a deterministic throw on valid-looking input, fixed by an actual `noexcept` signature. (The snippet also still calls `text.append`, so it doesn't remove the throwing growth path it claims to.) Bounding RSS and startup latency is the real win; I've written it up that way.

`fopen` also isn't "handled ✅". Two more paths, both measured:
- **FIFO** — `fopen` blocks forever, before the read loop is reached (rc=124 at a 10 s timeout). The host hangs in `rcl_node_init`.
- **Directory** — `fopen` succeeds, `fread` fails `EISDIR`, and `parseRateSpec("")` is *accepted* as a valid 0-topic/0-node spec. Alerting arms as a permanent no-op with no stderr output at all — the silent absence of monitoring this whole feature exists to prevent.

A 1 MiB cap fixes neither. Gating on `open(O_RDONLY|O_NONBLOCK)` + `fstat` + `S_ISREG` + `st_size` rejects oversize, char devices, FIFOs and directories in one pass, so that's what landed, in the pure core — which also means `tools/pulse_check.cpp:36-48`, which duplicates this loop verbatim, stops hanging on `--spec /dev/zero` in CI. Plus a warn-and-disable when a spec parses to zero rules.

Covered by a new `test/unit/test_spec_io.cpp` (7 cases incl. FIFO-without-blocking and `/dev/zero`) and integration cases for all three hostile paths. Good catch on the underlying issue.

TanayK07: Taking this, and it earns its keep beyond UX: while checking the read path I found an empty spec file — or a path that's a directory, where `fopen` succeeds and `fread` fails `EISDIR` — leaves `parseRateSpec` returning an accepted 0-topic/0-node spec with no stderr output whatsoever. Alerting looks armed and silently never fires, and `spec=0 topics, 0 nodes` is the only thing that would surface it.

So it landed together with a warn-and-disable for the zero-rule case. `m_spec` is populated in the constructor before `ensureStarted()` runs and the print is already behind the `m_started` CAS, so it's once per process image; `snprintf` into a fixed buffer keeps it allocation-free. Every disabling path — unset, unreadable, malformed, zero-rule — reports `spec=none`.

src/probe/interposers.cpp
TanayK07: Confirmed and fixed — though the mechanism isn't the one described, and the real one is worse.

The drain-before-exit story doesn't reproduce. The probe's atexit hook is registered at the *first tracepoint* (`:151`), so it runs **first** in the chain, not after rclcpp teardown: measured signal→tail-flush gap is 2.0–2.7 ms (SIGTERM and SIGINT alike), with the process taking a further ~112 ms to actually exit. So 40/3.5 = 11.43 Hz doesn't happen. Measured-window division also does its job for any tail ≥ 0.3 s — tail Hz came in at 50.34–51.07 against an interior of 50.00.

What actually breaks is the tail window's **length**. Over a sliver, `count/window_s` is a one- or two-sample estimate with ±1/window_s error **in both directions**, and exactly 0 Hz when the sliver caught nothing. Measured:

- `bench/stress_nodes pubfarm 5 30 8` at period=1.0 — executor self-cancels, no signal at all — leaves a 0.006–0.009 s zero-message tail and emits `WARN TOPIC /light_0 hz=0.000000 expected=[90,110]` on **3/3** runs, with `pulse-check` exit 1. At period=1.3 / 0.9 the tail is 1.110 s / 0.512 s and it's clean. So when run duration lands near a multiple of the flush period it's deterministic, not a flake.
- A 50 Hz talker with a 0.029 s tail reports **69.33 Hz** — breaking even the wide `[35,65]` spec our own integration test asserts green. Note the dominant failure is an *over*-rate warning, not the under-rate in the description.

On why the tests pass: partly as stated, but `run_single` also lands SIGTERM ~0.8–1.0 s into the tail with ±30% bounds, so quantization can't reach them; and the FAILING_SPEC test only asserts WARNs are *present*, which a bad tail satisfies too.

Your patch is the right fix and it's complete — the exit flush is the only source of a sub-period window, since periodic ones are ~period by construction and `forkChild` re-stamps the origin before re-arming. Landed as written, plus `test_exit_flush_window_is_not_alert_judged`, which asserts the invariant rather than racing the timing (min_hz 500 so every window violates regardless of length; verified red before, green after). Debounce deferred to R1.1 as you suggest — it'd cost ≥10 s of alert latency at the 5 s default and does nothing for pulse-check.

One thing I checked while in here and it's fine: `m_windows_flushed.fetch_add` sits *after* the empty-window early-out, so it really does count non-empty windows as its comment claims.

TanayK07: Confirmed — and docs-only isn't enough, so `--skip-last` landed.

`windows.back()` (`:134`) is always the exit-flush sliver for an exited process. It's actually worse than a flake in the `missing_as_zero` path: a fully-silent tail suppresses the TOPIC line entirely, so the spec topic is *absent* from the last window and gets scored at 0 Hz → `WARN TOPIC ... hz=0.000000` + exit 1. Measured rc=1 deterministically on the pubfarm log.

Two reasons documentation alone doesn't cover it: README.md:132-138 advertises exactly that post-run CI gate (`pulse-check --spec ... /tmp/pulse.*.log && echo healthy`), and the probe-side fix doesn't help here since pulse-check re-derives from raw rates and skips WARN lines by design.

Took your suggestion with one addition — it errors (exit 2) on a log with fewer than 2 windows rather than silently falling back to the ramp-up window, which the probe grace-skips for the same reason. Default behaviour is unchanged for live watchdogs, where the last window genuinely is current state. README now shows both forms and notes the ≥3-window guidance.

src/core/rate_spec.cpp
TanayK07: Verified against rclcpp source — you're right, and it's wider than described.

jazzy `publisher.hpp` L258-267 computes `inter_process_publish_needed = get_subscription_count() > get_intra_process_subscription_count() || buffer_`, and on the true branch calls `do_intra_process_ros_message_publish_and_return_shared` (fires `rclcpp_intra_publish`, L568) **and** `do_inter_process_publish` (fires `rcl_publish` via `rcl/src/rcl/publisher.c:282`). Both carry the same `rcl_publisher_t*`, so both land on one `sTopicCounter` and the sum is exactly 2×.

Two things that make it broader than "mixed subscribers":
- The intra tracepoint fires **unconditionally** at the top of the intra helper — it isn't gated on an intra subscriber existing. Since in-process subscriptions still create RMW endpoints, an IPC publisher whose subscribers are all *remote* hits this too.
- The `|| buffer_` term (jazzy+, `buffer_` created for TransientLocal durability) means **any** TransientLocal IPC publisher hits it regardless of subscribers. Iron's condition lacks that term.

`max` is right rather than "just read `rcl_publish`", exactly as you say — the all-in-process branch (L268-276) never calls rcl at all, so `pub_inter` is 0 there and reading it would report zero for a healthy publisher. I did consider rejecting `side: pub` + `transport: any` instead and decided against it: `any` is the default, so `{min_hz: 35, max_hz: 65, side: pub}` — the literal form in our own integration test — would become a parse error, and a parse error disables *all* alerting for the process. One over-strict rule silently killing the watchdog is a bad trade when there's a provably correct answer.

One correction: it's **Iron+**, not Jazzy+ — `rclcpp_intra_publish` is in ros2_tracing from iron onward. Humble confirmed unaffected both ways (no such tracepoint in its `tracetools.h`, and no tracepoint at all in its intra helpers), so the change is a no-op there. Our "jazzy+" wording elsewhere in the repo has the same off-by-one; left alone for now since Iron is EOL.

Landed with `RateSpecEval.PubAnyTakesMaxNotSumAcrossTransports` covering all three publish shapes plus a slow-publisher case (max mustn't rescue it) and an explicit recv/any assertion so nobody "fixes" the recv sum later. Verified red before — the (50,50) case reported `hz=100.000000`. README transport line and the `eRateTransport` docstring updated too.

TanayK07: Splitting this one off, because it isn't the same bug and I don't think `mergeStat` should change.

On the pub side one `publish()` increments two buckets — one physical event counted twice, wrong at any N. On the recv side each `callback_start` fires once per delivery and its `is_intra_process` flag selects exactly one bucket (`topic_registry.cpp:280-284`), so K subscriber processes are K genuinely distinct deliveries. That's a units question, not a double-count.

And it's already settled: `docs/issues/issue-1-double-count.md:76-77` explicitly ratified "total deliveries" as the intended per-topic recv semantic, as the documented consequence of the chosen option. Changing it in pulse-check would make offline and in-probe verdicts disagree. There's also no clean alternative available — the log carries per-topic aggregates, not per-endpoint counts, so pulse-check has no fan-out to normalize by even if it wanted to.

So: documented rather than changed. The pulse-check header and README now say plainly that recv is a delivery rate scaling with subscriber count, and that `max_hz` belongs on `side: pub`. Worth noting one follow-on from the `max` fix above — a cross-process merge where one process publishes intra-only and another inter-only now reports 50 rather than 100. Same corner case as mixed-IPC publishers in one process, and still strictly better than the 2× it replaces.

src/core/rate_spec.cpp
TanayK07: Verified the downstream half and took it.

`evaluateRateSpec` (`:255`) just walks the vector, and `observedHz` reads a different field per rule — I hand-built a two-sided spec against one stat with `pub_inter_hz=20, recv_inter_hz=3` and got exactly the one expected `WARN TOPIC /scan hz=3.000000 expected=[18,inf]`. `seen_topics` really is the only place uniqueness is assumed (whole-repo grep); the name-keyed containers in `pulse_check.cpp` and `log_reader.cpp` key on the *stats* side and are orthogonal. Ran your key against a 40-case grammar matrix: the only behavioural diffs are the intended ones, and the existing dupe regression at `test_rate_spec.cpp:107-108` stays failing unmodified, so it's strictly loosening.

Two notes on the framing. The KNOWN_ISSUES #1 citation is precedent for the buckets *existing* — it's about a counter double-count, not about operators wanting to spec both ends — so it argues feasibility rather than demand; I landed this on the merits instead. And `transport: any` is the inter+intra combining rule rather than a third bucket, so `{side: pub, transport: inter}` + `{side: pub, transport: any}` is redundant-but-satisfiable rather than contradictory — which is precisely why including transport in the key is safe. It can't detect genuine unsatisfiability (`{any, max_hz: 5}` + `{inter, min_hz: 10}`) and shouldn't try; satisfiability analysis has no business in a 200-line parser.

One thing your analysis surfaced indirectly: the dedup rigor was already uneven. `{min_hz: 1, min_hz: 2}` parsed silently as `min=2` — so "duplicate topics are hard errors" in the header was only half true. Closed that too, since it's the same defect class and two lines.

Landed here rather than as a follow-up given the rest of this batch already touches the grammar docs; header docstring and README updated with the two-sided form.

src/core/rate_spec.cpp
TanayK07: Confirmed, and the invisibility is as bad as you say. Reproduced:

```
BOM   -> line 1: unknown top-level entry '<EF BB BF>topics:' ...   (77 bytes)
TYPO  -> line 1: unknown top-level entry 'TOPICS:' ...             (74 bytes)
```

Nothing on screen distinguishes them — the only tell is the byte count. `nodes:` is affected identically.

Your CRLF read is right too: I checked `trim()` in all four positions (top-level key, flow-map body, `nodes: [...]` flow list, `- /a` block items) and CRLF is clean everywhere, so the BOM is the only Windows-artifact gap. Pinned all four in a test anyway, as you suggested.

One citation correction: Go is the wrong precedent — it's the ecosystem's deliberate holdout. `encoding/json` explicitly refuses to strip (golang/go#12254, closed unchanged; `encoding/csv` likewise per #9588/#48270), and Go's BOM handling is opt-in via `x/text`. Doesn't matter for the verdict — YAML 1.2.2 §5.2 makes `c-byte-order-mark` a grammar production consumed as document prefix, which is dispositive for something advertising itself as a YAML subset, and libyaml/PyYAML/SnakeYAML/.NET/Unicode 23.8.1 all check out. PyYAML's scanner is literally your one-strip-at-entry.

Also, the `string_view` sketch doesn't compile against the existing loop — `:153` does `std::string raw = text.substr(...)`, and `string_view::substr` returns a `string_view` whose `std::string` constructor is explicit. Went with `if (text.compare(0, 3, "\xEF\xBB\xBF") == 0) { pos = 3; }` after `:150`: one line, no new includes, safe on short input since `compare` clamps, and `line_no` still starts at 1 so error line numbers are unchanged. Confirmed the other 40 grammar cases are byte-identical.

Worth noting one caller boundary: pulse-check parses its spec through the same entry point, so it's covered for free. Its *log* reading path is separate and not covered — a BOM'd log silently drops the first window — but logs are probe-written via `fwrite` and never hand-authored, so I left that alone.

TanayK07:
Follow-up for anyone reading this thread later: the five review findings above were validated and fixed, but **not in this PR** — it was squash-merged from its original head (`348bb60`), so the fixes landed separately in #23 (`7e42c60`).

Where each one went:

| Finding | Verdict | Landed in |
|---|---|---|
| Unbounded spec-file read | Partly — see below | #23, as an `S_ISREG` + `O_NONBLOCK` + 1 MiB gate |
| Exit-time partial window WARNs | Confirmed | #23, `flush(exiting)` + `pulse-check --skip-last` |
| `transport: any` pub double-count | Confirmed | #23, `max` for pub, sum kept for recv |
| One rule per topic | Confirmed | #23, dedup on (topic, side, transport) |
| UTF-8 BOM | Confirmed | #23, one strip at parse entry |

Two corrections to what I wrote above, for the record: the spec-read issue is a **resource bound**, not a `std::terminate` risk (`extern "C"` governs linkage, not unwinding, and the constructor already allocates in four other places) — and the exit-window failure is **not** the drain-before-exit gap I described. The probe's atexit hook runs first in the chain, so the measured gap is 2.0–2.7 ms; what actually breaks is the sliver window's length. Details in #23.

Also found while checking these: a directory or empty spec file parsed as a valid **zero-rule spec**, arming alerting as a permanent silent no-op with no stderr at all. Fixed in #23 too.

Shipped in `v0.2.0`.

PR #24, opened 2026-08-02
docs: R5 implementation spec, roadmap rescope (R5 promoted, R2 corrected)

TanayK07:
Two docs commits on the roadmap.

**1. Stray `Jazzy+`** on the LTTng-coexistence item, leftover from #23's Iron+ correction, and it was never a tracepoint-availability claim anyway. Reworded to name what actually gates it (binaries carrying the lttng-ust backend).

**2. R5 promoted, R2 rescoped**, from researching R2 before starting it.

R2's justifying case doesn't hold: the canonical ROS 2 control loops aren't rclcpp timers. `ros2_control`'s `ros2_control_node` runs a raw `std::thread` with `sleep_until`, Nav2's `controller_server` uses `nav2::Rate`, MoveIt Servo uses `rclcpp::WallRate`, none emit timer tracepoints. So "catches a wedged control loop" is precisely what R2 can't do. And `ros2_control` on Jazzy+ already ships periodicity avg/min/max/stddev with an overrun count.

R5 turns out to matter more, because it closes a soundness hole in the R1 alerting just shipped in v0.2.0: a windowed mean can't see a stall. At 50 Hz with `min_hz: 45`, a 400 ms freeze reports 46 Hz and passes, and widening the window makes it *worse*, not better. Max gap is the window-length-independent detector, and it covers the `Rate`-driven loops R2 misses via the topics they publish.

Corrections to R2, all verified against upstream source rather than assumed:

- **No per-fire timer tracepoint exists.** The per-fire signal is `callback_start` from `GenericTimer::execute_callback()`, which the probe already interposes, timer fires reach `onCallbackStart` today and hit the `kNotASubscription` sentinel. R2 is three init-time hooks, not a new counting path.
- **"Retires the negative-cache sentinel" struck**, it's a net cost (early return becomes a `fetch_add`), the sentinel can't be retired regardless, and `issue-3-callback-lock-storm.md` already rejected eager timer resolution.
- `TIMER /node period_ms=20 actual_hz=49.8` has **no unique key**; declared period is **not** a usable expected rate (sim time, `exchange_period`, `cancel` all untraced); **rclpy timers orphan**; **services cut** (6 parameter services per node).
- "All exist on Humble" **was correct**, and they exist on iron/jazzy/kilted/rolling too, identical signatures.

What survives of R2 is liveness, not rate: it fixes the documented gap where a pure timer/service node with no topic traffic reads as quiet.

Docs-only. R5's concrete design lands separately.


TanayK07:
Added `docs/design-r5-gap-visibility.md` — the full implementation spec for R5, plus a ROADMAP pointer to it.

Everything in it is measured on this tree or cited to `file:line`. Highlights:

**Cost is fine.** +24.4 ± 0.02 ns/msg, paired order-alternated N=15. The ROADMAP's "~20 ns" estimate was right, and ~96% is the vDSO clock read — so the accumulation design has no performance content and gets chosen on correctness. That's 0.012% of a core at 4900 msg/s, i.e. **1/15th of one SEM** of our own harness. Resolving it end-to-end would take ~6400 paired trials, so the docs should publish the microbench delta and the arithmetic rather than invent an end-to-end number.

**Three departures from the sketch:**
- `min_dt_ms` **cut** — the Iron+ double-fire from #23 puts two timestamps ~1 µs apart for one `publish()`, so it reads ~0.001 ms on a healthy topic forever.
- **Both sides**, not RECV-only — rclpy is publish-side-only visible, so RECV-only is blind to every Python publisher.
- **New `JITTER` line**, not appended fields — `probe_harness.py:32-34` anchors with `$`, so appending silently breaks nine integration tests. Verified.

**Two holes found in R5 as specced**, both load-bearing: a fully-dead endpoint produces no inter-arrival pair, so the detector misses the *total* stall unless flush folds in `now - last_ts`; and that fold must be suppressed in the exit window or healthy teardown fires every gap rule deterministically.

**One prerequisite, independently justified:** 63 of 128 `sTopicCounter`s currently have their count-atomics straddling a cache line — a latent repeat of KNOWN_ISSUES #15, measured at 4.5× on neighbouring endpoints under 8 threads. `alignas(64)` + hot-field reorder lands as its own `perf:` commit before R5, not bundled.

Still docs-only. Implementation is three commits after this merges.

src/core/topic_registry.cpp
```diff
@@ -317,6 +393,16 @@ auto TopicRegistry::snapshot(double window_s) -> std::vector<sTopicStat> {
         s.recv_inter_hz = static_cast<double>(ri) / w;
         s.recv_intra_hz = static_cast<double>(rx) / w;
         s.recv_endpoint_seen = c->recv_endpoint_seen;
+        if (gap) {
+            // Reset the per-window max but NOT last_*: the next window's first message must
+            // measure its gap back into this one, or a stall straddling the flush is lost.
+            // Folding the still-open interval is what makes a fully-silent endpoint report a
+            // GROWING number instead of dropping out of the report entirely.
+            s.has_pub_max_dt = takeGap(c->pub_last_ns, c->pub_max_dt_ns, now_ns, fold_open_gap,
```
TanayK07: ## 🔬 ISSUE: filtered topics never reset their gap accumulators

**Severity:** MEDIUM — unbounded stale state on `/rosout`, `/parameter_events`, `/diagnostics`, and a wrong first reading if the filter list ever changes.

### The Problem

These `takeGap()` calls sit **after** the `shouldFilter()` `continue`, so a filtered topic never drains `pub_max_dt_ns` / `recv_max_dt_ns`. They ratchet upward for the whole process lifetime.

### Theory / Proof

The ordering directly above is not accidental. `snapshot()` exchanges all four count buckets *before* the filter check, and the comment there says why: *"Exchange all three split buckets first (this also resets them), so filtered topics still count toward node activity even though they are excluded from the stats output."* Whoever wrote that hit this exact hazard and solved it by resetting before filtering. The R5 code does not follow the pattern it is nested inside.

This is not a theoretical object either: `linkNodeCounter` links filtered counters into `sNode::counters`, and `/rosout` is high-traffic on any real graph, so these are long-lived and hot.

Invisible today — filtered topics are never emitted and never spec-matched, so the stale value is written and never read. Two ways it becomes real:

1. `shouldFilter` is a hardcoded list of three topics. Make it configurable, or let a spec assert `/diagnostics` liveness, and the first reported window emits a `max_dt_ms` accumulated **since process start** rather than since the last window — an hours-old number presented as this window's gap.
2. It is a silent divergence from a reset discipline the surrounding code documents, which is the kind of thing that survives until it bites.

### Research

Same class as the counter-reset ordering established in `docs/issues/issue-1-double-count.md`: per-window state must be drained on a path that cannot be skipped by a reporting filter, because "not reported" and "not accumulated" are different properties.

### Recommendation

Drain in the same block as the counters, keep only the reporting after the filter:

```cpp
uint64_t rx = c->recv_intra.exchange(0, std::memory_order_relaxed);
double pub_gap_ms = 0.0, recv_gap_ms = 0.0;
bool has_pub_gap = false, has_recv_gap = false;
if (gap) {
    has_pub_gap  = takeGap(c->pub_last_ns,  c->pub_max_dt_ns,  now_ns, fold_open_gap, pub_gap_ms);
    has_recv_gap = takeGap(c->recv_last_ns, c->recv_max_dt_ns, now_ns, fold_open_gap, recv_gap_ms);
}
if (p > 0 || px > 0 || ri > 0 || rx > 0) { active.insert(c); }
if (shouldFilter(c->topic)) { continue; }
...
s.has_pub_max_dt = has_pub_gap;   s.pub_max_dt_ms  = pub_gap_ms;
s.has_recv_max_dt = has_recv_gap; s.recv_max_dt_ms = recv_gap_ms;
```

Add a unit test asserting a filtered topic's `max_dt_ns` is zero after `snapshot()` — nothing observable would otherwise catch a regression.

src/probe/interposers.cpp
```diff
@@ -260,7 +282,11 @@ class ProbeRuntime {
         const auto now_mono = std::chrono::steady_clock::now();
         const double window_s = std::chrono::duration<double>(now_mono - m_window_start).count();
         m_window_start = now_mono;
-        auto stats = m_registry.snapshot(window_s);
+        // Gap tracking (R5): fold the still-open interval into each endpoint's reported
+        // max, EXCEPT on the exit window — rclcpp teardown stops traffic before the
+        // process exits, so the open gap there measures the shutdown sequence and would
+        // fire every max_gap_ms rule on a perfectly healthy stop.
+        auto stats = m_registry.snapshot(window_s, /*fold_open_gap=*/!exiting);
```
TanayK07, line 289: ## 🔬 GAP ANALYSIS: R5 ships with no integration test, including the one that proves the feature

**Severity:** HIGH — the central claim of the feature is unverified end-to-end, and the design doc names the test that would verify it.

### The Problem

`git diff origin/main --stat -- test/integration/` is empty. R5 adds 11 unit tests and zero integration tests.

### Theory / Proof

The unit tests exercise `TopicRegistry` directly with synthetic handles and `sleep_for`. None of them run the probe against a real ROS graph, so nothing verifies:

- that `callback_start` / `rcl_publish` reach `noteArrival` through the actual LD_PRELOAD path,
- that the emitted `JITTER` line survives `parseLog` → `mergeStat` → `evaluateRateSpec` in `pulse-check`,
- that the exit-window fold suppression works against **real** rclcpp teardown, which is the thing it exists to defend against. `GapTracking.OpenGapNotFoldedWhenExiting` passes `fold_open_gap=false` by hand; it does not prove `flush(exiting=true)` threads it through.

Most importantly, the R5 thesis itself is untested. `docs/design-r5-gap-visibility.md` specifies it by name:

> **`test_gap_detected_where_mean_rate_passes`** — the R5 thesis as an executable test, and the most valuable one here. 50 Hz, period 5 s, run 16 s, 800 ms stall at t=7 s. The stalled window measures ~42 Hz. Spec `{min_hz: 40, max_gap_ms: 100, side: pub}`. Assert **zero** `WARN TOPIC /chatter hz=` anywhere *and* ≥1 `max_dt_ms` WARN ≥ 800.

Without it, "a windowed mean cannot see a stall but max_dt_ms can" is an argument in a commit message, not a property the suite defends.

The manual end-to-end checks in the commit message did pass, so the feature works right now. But a refactor that resets `last_ts` in `snapshot()`, or drops the fold, or breaks the pulse-check merge, would leave all 103 unit tests green — the exact regressions the design flagged as load-bearing are the ones with no automated coverage.

### Research

The design doc also specifies `stall_pub <stall_ms> <at_s>` in `test_nodes.cpp` and explains why it is deterministic rather than racy: `sleep_for` inside the callback blocks the single-threaded executor, so `max_dt_ms >= stall_ms` is a hard one-sided guarantee, and because `last_ts` survives the flush boundary the full dt lands in one window however the stall straddles it. That is a materially better pattern than `test_stall.py`, which kills a process and races the flush.

### Recommendation

Add `test/integration/test_jitter.py` plus the `stall_pub` node mode, with at minimum:

```python
def test_gap_detected_where_mean_rate_passes():
    """The R5 thesis: mean rate passes, gap rule catches it."""
    spec = "topics:\n  /chatter: {min_hz: 40, max_gap_ms: 100, side: pub}\n"
    text = ph.run_single("stall_pub", out, run_s=16.0, period="5.0",
                         extra_env={"ROS_TOPIC_STATS_EXPECTED": spec})
    assert not [w for w in _warn_lines(text) if " hz=" in w], "mean rate must NOT fire"
    gaps = [w for w in _warn_lines(text) if "max_dt_ms=" in w]
    assert gaps and float(re.search(r"max_dt_ms=([\d.]+)", gaps[0]).group(1)) >= 800
```

and `test_exit_window_gap_not_folded`, which is the only way to prove the `flush(exiting)` plumbing rather than the `snapshot()` parameter.

tools/pulse_check.cpp
```diff
@@ -172,8 +191,21 @@ int main(int argc, char** argv) {
         }
     }
 
-    const auto warnings =
-        evaluateRateSpec(*spec, stats, active_nodes, known_nodes, /*missing_as_zero=*/true);
+    std::vector<std::string> unmeasured;
+    const auto warnings = evaluateRateSpec(*spec, stats, active_nodes, known_nodes,
+                                           /*missing_as_zero=*/true, &unmeasured);
+    // A gap rule the logs cannot answer is bad input, not a verdict: exit 0 would claim it was
+    // checked and healthy, exit 1 would send someone chasing a stall that was never measured.
```
TanayK07, line 198: ## 🔬 ISSUE: `pulse-check` exits 2 before printing real violations it already computed

**Severity:** MEDIUM — a spec with one unmeasurable gap rule silently hides every genuine failure in the same run.

### The Problem

```cpp
const auto warnings = evaluateRateSpec(..., &unmeasured);
if (!unmeasured.empty()) {
    for (const auto& t : unmeasured) { std::fprintf(stderr, "..."); }
    return 2;                      // <-- warnings computed, then discarded
}
for (const auto& w : warnings) { std::printf("%s\n", w.c_str()); }
```

`warnings` is fully populated on the line above and then thrown away. The early return happens whenever **any** rule has an unmeasured gap, regardless of what else failed.

### Theory / Proof

Concrete: a spec with ten topics, nine `min_hz` rules and one `max_gap_ms` rule. The probe ran without jitter, so the gap rule is unmeasurable — but three of the nine rate rules are violated and `evaluateRateSpec` has already produced their `WARN` lines. The operator sees only:

```
pulse-check: spec sets max_gap_ms for /scan but the logs carry no JITTER line for it
```

exit 2. Three real, measured, currently-failing rate checks are suppressed by a configuration problem on an unrelated topic. In a CI gate that reads as "bad input, fix your invocation" — so someone fixes the env var, reruns, and only then discovers the stack was broken the whole time.

The exit-2 *code* is right: exit 0 would claim a check happened, exit 1 would send someone chasing a stall that was never measured. That reasoning is sound and should stay. The bug is discarding orthogonal, already-computed results — measurement absence on topic A says nothing about a measured violation on topic B.

### Research

`ros2 doctor` and `systemd-analyze verify` both report all findings and set the exit code from the most severe. `shellcheck` likewise emits every diagnostic and exits non-zero once. Reporting-completeness and exit-code severity are independent axes; collapsing them loses information the tool already has in hand.

### Recommendation

Print everything, then pick the exit code by severity:

```cpp
for (const auto& w : warnings) {
    std::printf("%s\n", w.c_str());
}
if (!unmeasured.empty()) {
    for (const auto& t : unmeasured) {
        std::fprintf(stderr,
                     "pulse-check: spec sets max_gap_ms for '%s' but the logs carry no JITTER "
                     "line for it — was the probe run with ROS_TOPIC_STATS_JITTER=1?\n",
                     t.c_str());
    }
    return 2;  // unmeasurable dominates: the verdict is incomplete either way
}
return warnings.empty() ? 0 : 1;
```

Worth an integration case asserting stdout carries the rate WARN while the exit code is still 2.

README.md
```diff
@@ -78,8 +78,13 @@ cd ~/ros2_ws/src && git clone https://github.com/TanayK07/ros2_pulse.git
...
 
-Requires: ROS 2 Humble, a `libtracetools.so` with instrumentation compiled in (the default on
-Humble/Isaac binaries — verify with `nm -D $(ros2 pkg prefix tracetools)/lib/libtracetools.so* | grep -c ros_trace`).
+Requires ROS 2 **Humble, Jazzy or Kilted** (all three CI-tested on stock `ros:<distro>` images —
+see [Compatibility](#compatibility)) and a `libtracetools.so` with instrumentation compiled in,
+which is the default on stock and Isaac binaries. Verify with:
+
+```bash
+nm -D $(ros2 pkg prefix tracetools)/lib/libtracetools.so* | grep -c ros_trace
+```
```
TanayK07, line 88: ## 🔬 GAP ANALYSIS: R5 is completely undocumented for operators

**Severity:** HIGH — a shipped feature with no README presence, in a project whose stated failure mode is silent absence of monitoring.

### The Problem

`grep -c "JITTER\|ROS_TOPIC_STATS_JITTER\|max_gap_ms" README.md` returns **0**. The PR adds:

- an env var, `ROS_TOPIC_STATS_JITTER`
- a new output line kind, `JITTER <topic> <side> max_dt_ms=…`
- a new spec key, `max_gap_ms`
- a new WARN line, `WARN TOPIC … max_dt_ms=… expected_max_gap_ms=…`
- a new `pulse-check` exit-2 condition
- an implicit-enable rule (a `max_gap_ms` spec turns tracking on)

None of it appears in the README, which documents every other env var in a table at lines 100-108, shows the window format at lines 40-51, and specifies the spec grammar at 106-160.

### Theory / Proof

This matters more here than in a typical project, for two reasons.

First, R5 exists *because* R1's alerting silently misses stalls. An operator who has read the README believes `min_hz` protects them. Nothing tells them a 400 ms freeze at 50 Hz passes a `min_hz: 45` rule, or that `max_gap_ms` is the fix. The feature that closes the hole is invisible to exactly the person holding the hole.

Second, the project treats undocumented-but-active behaviour as a defect it fixes deliberately: the zero-rule-spec warning, the `spec=` banner field, and the `--skip-last` README block all exist because "why am I getting no WARNs?" was judged unacceptable. Shipping `jitter=on/off` in the banner with no README entry explaining it repeats that.

Today: nobody can discover the feature except by reading the diff. The `pulse-check` exit-2 message is the only operator-facing string that mentions `ROS_TOPIC_STATS_JITTER`, and it is reachable only *after* you have already written a `max_gap_ms` rule you could not have known about.

### Recommendation

Four edits, all in existing structures:

1. Env table (after line 108):
   `| ROS_TOPIC_STATS_JITTER | 0 | When 1, measure per-endpoint inter-arrival gaps and emit a JITTER line per side. Also enabled implicitly by any max_gap_ms rule in the spec. Costs ~24 ns per message (one clock read); default off. |`
2. Output block (lines 40-51): add a `JITTER /points pub max_dt_ms=20.860` line.
3. Expected-rate section: add `max_gap_ms` to the grammar example and state plainly *why* — a windowed mean cannot see a stall; give the 400 ms / 46 Hz arithmetic.
4. `pulse-check` block: note the exit-2 case for a gap rule the logs cannot answer.

`CHANGELOG.md`'s `[Unreleased]` section is also empty and should carry these entries — v0.2.0 is already tagged, so R5 belongs there.

src/core/topic_registry.cpp
```diff
@@ -317,6 +393,16 @@ auto TopicRegistry::snapshot(double window_s) -> std::vector<sTopicStat> {
...
+                                       s.pub_max_dt_ms);
+            s.has_recv_max_dt = takeGap(c->recv_last_ns, c->recv_max_dt_ns, now_ns,
+                                        fold_open_gap, s.recv_max_dt_ms);
+        }
```
TanayK07: ## 🔬 ISSUE: gap state survives `fork()`, so the child inherits the parent's timeline

**Severity:** LOW — a bounded, one-window artifact, but it is a new instance of a hazard this repo has a documented policy for.

### The Problem

`forkChildReset()` re-initializes `m_mu` and nothing else. The R5 fields — `pub_last_ns`, `recv_last_ns`, `pub_max_dt_ns`, `recv_max_dt_ns` — are inherited verbatim by the child.

### Theory / Proof

Two distinct behaviours, and they differ:

- **`last_ns` inheriting is arguably correct.** `CLOCK_MONOTONIC` is per-boot and shared across processes, so `now - parent_last_ns` in the child is a real elapsed interval. A fork-without-exec child that continues publishing measures a meaningful first gap.
- **`max_dt_ns` inheriting is not.** The child's first window reports the largest gap the *parent* saw, attributed to the child. With the open-gap fold on, a child forked from a parent whose topic had been quiet reports that quiet as its own.

This is the same smear `docs/issues/issue-10-fork-without-exec.md` documents for counts — *"Inherited pre-fork counts may smear into the child's first window"* — but the counts case was reasoned about and written down, and this one is neither reset nor documented.

Requires `fork()` without `exec()` in a probed process with jitter enabled — narrow, which is why this is LOW. When it happens, the child's first window can emit a `max_dt_ms` that is entirely the parent's, and with a `max_gap_ms` rule that is a false alarm on a healthy child. Self-correcting after one window, since `max_dt_ns` resets in `snapshot()`.

### Research

`pthread_atfork` child handlers conventionally reset per-process derived state while leaving process-independent clocks alone — which is exactly the split here: keep `last_ns`, drop `max_dt_ns`.

### Recommendation

The child is single-threaded at this point, so a plain loop is safe:

```cpp
void TopicRegistry::forkChildReset() {
    new (&m_mu) std::shared_mutex();
    // Gap state (R5): last_ns is CLOCK_MONOTONIC and process-independent, so the child's first
    // interval measured against it is real and is kept. The per-window maxima are the PARENT's
    // observations, so drop them rather than attribute them to the child's first window.
    for (auto& kv : m_by_topic) {
        kv.second->pub_max_dt_ns.store(0, std::memory_order_relaxed);
        kv.second->recv_max_dt_ns.store(0, std::memory_order_relaxed);
    }
}
```

If you would rather not add work to the fork path, the alternative is a sentence in `docs/issues/issue-10-fork-without-exec.md` extending the documented smear to gaps — but silent is the one option that does not fit this repo.

src/core/topic_registry.cpp
```diff
@@ -50,6 +51,54 @@ inline auto tlsSlotFor(const void* key) -> sTlsSlot& {
...
+inline void noteArrival(std::atomic<uint64_t>& last_ns, std::atomic<uint64_t>& max_dt_ns) {
+    const auto now = static_cast<uint64_t>(
+        std::chrono::steady_clock::now().time_since_epoch().count());
+    const uint64_t prev = last_ns.exchange(now, std::memory_order_relaxed);
+    if (prev == 0 || now <= prev) {
+        return;  // first arrival ever, or a reordered pair — no usable interval
+    }
+    const uint64_t dt = now - prev;
+    // std::atomic::fetch_max is C++26; this is the portable form. It costs nothing in steady
+    // state because new maxima follow the harmonic number — ~14 CAS per 500 messages.
```
TanayK07, line 76: ## 🔬 ISSUE: a single-message window reports `max_dt_ms=0.000`, which reads as "perfectly regular"

**Severity:** LOW — misleading value in the exit window, and the pinned test encodes the behaviour rather than questioning it.

### The Problem

`takeGap()` gates on `last == 0` (never active), not on whether an actual **interval** exists. After exactly one arrival, `last_ns != 0` but no inter-arrival pair has been observed, so with `fold_open_gap=false` the function returns `true` with `dt == 0` and the block emits:

```
JITTER /chatter pub max_dt_ms=0.000
```

`GapTracking.FirstArrivalProducesNoSample` asserts exactly this — `EXPECT_TRUE(s->has_pub_max_dt)` with `EXPECT_DOUBLE_EQ(s->pub_max_dt_ms, 0.0)` — so the test name says "produces no sample" while the assertions require it to report one.

### Theory / Proof

`0.000` is not a neutral value for this field: it is the *best possible* reading, indistinguishable from a topic whose arrivals are perfectly spaced. The design doc is explicit that this distinction matters — for the never-active case it argues *"Reporting 0.000 would read as 'perfectly regular'"* — but the one-arrival case has the same property and is not covered.

Reachability is narrow because the production path folds the open gap, which turns the single-arrival case into a real `now - last_ts`. `fold_open_gap=false` is used in exactly one place: the exit flush. So this surfaces on a topic that published once and then the process exited.

Today: a topic with one message in its lifetime shows `max_dt_ms=0.000` in the final window. Harmless for alerting, since `0 > max_gap_ms` is false — but it is a false negative by luck, not by design, and anyone reading the log sees a claim of perfect regularity from a single data point. If a future rule ever gains a *minimum*-gap or regularity check, it silently passes.

### Recommendation

Distinguish "has a timestamp" from "has an interval". A second sentinel is not needed — a zero `max_dt_ns` with the fold disabled already means "no interval observed":

```cpp
inline auto takeGap(std::atomic<uint64_t>& last_ns, std::atomic<uint64_t>& max_dt_ns,
                    uint64_t now_ns, bool fold_open_gap, double& out_ms) -> bool {
    const uint64_t last = last_ns.load(std::memory_order_relaxed);
    if (last == 0) { return false; }
    uint64_t dt = max_dt_ns.exchange(0, std::memory_order_relaxed);
    if (fold_open_gap && now_ns > last) {
        const uint64_t open = now_ns - last;
        if (open > dt) { dt = open; }
    }
    if (dt == 0) {
        return false;  // a timestamp but no measured interval — not the same as a zero gap
    }
    out_ms = static_cast<double>(dt) / 1e6;
    return true;
}
```

The only sample this discards is a genuine `dt == 0` (two arrivals inside one ~21 ns clock tick), which can never be a maximum and is already dropped by `noteArrival`'s `now <= prev` guard. Then rename the test to match what it asserts — `FirstArrivalProducesNoInterval`, expecting `has_pub_max_dt == false`.

src/core/topic_registry.cpp
TanayK07: Fixed in `3cbc38b`. Confirmed the read: `takeGap()` sat after the `shouldFilter` `continue`, so `/rosout` and friends never drained `max_dt_ns`. Both calls now happen in the same block as the count exchanges, before the filter, and only the assignment to `sTopicStat` stays after it — matching the discipline the comment on those exchanges already documented.

Added `GapTracking.FilteredTopicsStillDrainTheirGapState` so a regression has something to trip over, though the honest caveat is that a filtered topic is unobservable by construction, so the test asserts the proxy (it stays absent from the snapshot) rather than the drained value itself.

src/probe/interposers.cpp
TanayK07: Fixed in `3cbc38b` — this was the right call to make loudest.

Added `test/integration/test_jitter.py` (6 cases) and the `stall_pub` node mode from the design doc: a 20 ms wall-timer talker that calls `sleep_for` **inside** the callback, so the single-threaded executor blocks for exactly that long. Deterministic, unlike `test_stall.py`'s process-kill, and it does not care where the flush boundary lands because `last_ts` survives it.

`test_gap_detected_where_mean_rate_passes` is the one that matters: 50 Hz, 5 s windows, 800 ms freeze at t=7. The stalled window measures ~42 Hz so `min_hz: 40` stays green, while `max_gap_ms: 100` fires at >=800 ms. It also guards its own premise — if the rate rule ever fires, the test fails with "this stall was NOT invisible to min_hz" rather than passing for the wrong reason.

`test_exit_window_gap_is_not_folded_or_judged` covers your specific point about `flush(exiting=true)`: the unit test only proved the `snapshot()` parameter, this proves the plumbing.

144 tests, 0 failures.

tools/pulse_check.cpp
TanayK07: Fixed in `3cbc38b`. You are right that the two axes are independent — measurement absence on topic A says nothing about a measured violation on topic B.

Violations now print before the unmeasured check; exit 2 still wins, since the verdict is incomplete either way. Kept the exit-2 *code* for the reasons in the original comment (exit 0 would claim a check happened, exit 1 would send someone chasing a stall nobody measured) — the bug was only in throwing away results already in hand.

Pinned by `test_pulse_check_reports_measured_violations_alongside_exit_2`: a spec with one violated `min_hz` rule and one unmeasurable `max_gap_ms` rule must produce the rate WARN on stdout **and** exit 2.

README.md
TanayK07: Fixed in `3cbc38b`. `grep -c "JITTER\|max_gap_ms" README.md` was genuinely 0 — the whole feature was diff-only.

Added all four: the `ROS_TOPIC_STATS_JITTER` env row (with the measured +24 ns and the implicit-enable note), a `JITTER` line in the output block, a **"Catching stalls a rate bound cannot see"** subsection carrying the arithmetic — 225-message threshold, 400 ms freeze reads 46 Hz, widening the window makes it worse — and the `pulse-check` exit-2 caveat. Plus a `CHANGELOG [Unreleased]` entry, since 0.2.0 is already tagged.

The subsection also notes the `Rate`-driven case (ros2_control, Nav2, MoveIt Servo), because that is the practical reason someone would reach for a gap bound over a rate bound and it was nowhere in the operator docs.

src/core/topic_registry.cpp
TanayK07: Fixed in `3cbc38b`, taking the reset — but the tension you flagged with issue #10 is real and worth stating in the code, so the comment now says why this diverges.

`forkChildReset()` clears `pub_max_dt_ns` / `recv_max_dt_ns` and keeps `last_ns`. The split is the one you described: `CLOCK_MONOTONIC` is per-boot and process-independent, so an interval the child measures against the inherited timestamp is real, while the maxima are the parent's observations.

The reason to diverge from the documented counts smear rather than just extend the doc: a stale count cannot raise an alert, a stale gap can — a child forked from a parent whose topic had been quiet would fire `max_gap_ms` on its first window while being perfectly healthy. That asymmetry is what makes it worth the loop, and the child is single-threaded at that point so the loop is safe.

src/core/topic_registry.cpp
TanayK07: Fixed in `3cbc38b`. Agreed the test name was arguing against its own assertions.

`takeGap` now returns false when `dt == 0`, distinguishing "has a timestamp" from "has an interval". The only genuine sample that discards is two arrivals inside one ~21 ns clock tick, which can never be a maximum and is already dropped upstream by `noteArrival`'s `now > prev` guard.

Renamed to `FirstArrivalProducesNoInterval` and flipped the assertion, then extended it to cover the other half: with the open gap folded (the production path) the same single arrival *does* become a real measurement. That pairing is what makes the narrowness explicit — the 0.000 was only ever reachable in the exit window.

Fixing this also invalidated `OpenGapNotFoldedWhenExiting`, which had been asserting the 0.000 behaviour. Rewrote it to publish twice for a real small interval and then idle 60 ms, so it now tests what it always meant: the measured interval is reported and the teardown quiet is not.

PR #25, opened 2026-08-08
test+bench: close the R5 test-plan gaps: race stress, reader locks, reproducible +24 ns figure

Tanay10x:
## What

R5 (#24) shipped the gap-visibility feature but skipped four items its own design doc (`docs/design-r5-gap-visibility.md`, "Test plan") calls for. This PR closes them, test-first, with mutation evidence that each new test has teeth:

- **`ConcurrencyStress.JitterMaxIsRaceFree`**, 8 threads hammer one endpoint per side around a known 25 ms stall. Two-sided assertion: the stall must be visible (≥25 ms) and `max_dt` can never exceed the test's wall time. The design doc documents the failure mode this guards (unguarded reordered exchange → unsigned underflow → ~1.8e19 ns poisons `max_dt` permanently); **mutation check: removing the `now > prev` guard fails 3/3 runs at exactly 1.84467e+13 ms.** TSan clean over 5 repeats.
- **Log-reader JITTER locks**, round-trip of values AND `has_*` flags (the flag drives pulse-check's exit-2 "cannot answer" path, dropping it would turn measurement absence into a health verdict), pre-R5 logs parse with flags false, unknown JITTER side is skipped without conjuring a topic entry.
- **`DisabledPathNeverEntersTheAccumulator`** (the plan's `DisabledReadsNoClock`), implemented without a clock seam: `noteArrival` writes `last_ns` unconditionally *before* its guard, so traffic-while-disabled + enable + fold-armed snapshot reports a gap iff the accumulator was ever entered, and its first statement is the clock read. **Mutation check: removing the `if (gap)` gate fails the test.**
- **Bench gap leg**, the published "+24.4 ns/msg (96% clock read)" figure had no committed harness. `hotpath_bench` now models the *current* 256-slot stride-hash cache (it still labeled the pre-#15 16-slot design "current"), adds a `+gap` twin running an exact `noteArrival` replica, a raw-clock attribution leg, and prints the delta in **CPU ns/msg** (the leg convention amortizes wall time across threads, which divides per-message cost by parallelism). Measured here: **+23.7 CPU ns/msg fixed / +25.3 alt-4, clock alone 22.1 CPU ns = 93%**, reproducing the published figure.

## Verification

- fast lane: 109/109
- TSan lane: clean (`setarch -R` needed locally on kernel 6.8; CI containers unaffected)
- both mutations verified red before landing green


test/unit/test_gap_tracking.cpp
```diff
@@ -203,6 +203,38 @@ TEST(GapTracking, OpenGapFoldedSoSilentEndpointGrows) {
...
+    // Flip on AFTER the traffic; the snapshot drains with the fold armed. A regression that
+    // calls noteArrival unconditionally (e.g. "simplifying" the if (gap) gate away) leaves a
+    // last_ns behind, and this fold turns it into a visible JITTER report.
+    reg.setGapTracking(true);
+    auto stats = reg.snapshot(1.0, /*fold_open_gap=*/true);
+    const auto* s = findStat(stats, "/chatter");
+    ASSERT_NE(s, nullptr);
+    EXPECT_EQ(s->pub_inter_count, 2u);  // counting itself was live the whole time
+    EXPECT_FALSE(s->has_pub_max_dt) << "disabled phase left a timestamp -> a clock was read";
+    EXPECT_FALSE(s->has_recv_max_dt);
```
TanayK07, line 235: ## 🔬 ISSUE — the disabled-path proof covers only the publish side; the receive-side gates survive the exact mutation it names

**Severity:** HIGH — with both `onCallbackStart` gap gates deleted, the whole suite stays green at 109/109. Every callback delivery would pay the R5 clock read with `ROS_TOPIC_STATS_JITTER` off, and nothing catches it.

### The problem

`EXPECT_FALSE(s->has_recv_max_dt)` on this line is vacuous. The test registers a publisher and nothing else:

```cpp
reg.onNodeInit(kNode, "talker", "/");
reg.onPublisherInit(pub, kNode, "/chatter");
reg.onPublish(pub);   // ... and no onCallbackStart, ever
```

`recv_last_ns` is written only by `noteArrival` from `onCallbackStart` (`src/core/topic_registry.cpp:321`, `:361`). With no subscription and no delivery it is `0`, so `takeGap` returns `false` at `topic_registry.cpp:88` regardless of what the receive-side gate does. The assertion passes for an implementation with no receive gate at all.

### Theory / proof

The gate exists at three call sites, not one:

| Site | `topic_registry.cpp` | Covered by this test? |
|---|---|---|
| `publishCount`, TLS hit | 288-290 | yes |
| `publishCount`, shared-lock miss | 299-301 | yes |
| `onCallbackStart`, TLS hit | 320-322 | **no** |
| `onCallbackStart`, shared-lock/resolve | 360-362 | **no** |

`test/unit/test_gap_tracking.cpp` never calls `onCallbackStart` anywhere in the file — `grep -c onCallbackStart test/unit/test_gap_tracking.cpp` is 0. `test/stress/test_concurrency.cpp:121` does drive callbacks, but with `setGapTracking(true)`, so removing the gate changes nothing there. Every other registry test leaves tracking off, and `snapshot()` then short-circuits at `topic_registry.cpp:370` (`gap == false` ⇒ `takeGap` is never called), so a stray `recv_last_ns` is unobservable.

### Mutation evidence — run on this branch at `cd67b45`

Both receive-side gates removed, i.e. `noteArrival(ctr->recv_last_ns, ctr->recv_max_dt_ns);` called unconditionally at `:320` and `:360`:

```
[==========] 109 tests from 19 test suites ran. (1771 ms total)
[  PASSED  ] 109 tests.
```

The publish-side gate removed instead (`if (gap)` dropped in `publishCount`):

```
[  FAILED  ] GapTracking.DisabledPathNeverEntersTheAccumulator (5 ms)
[  PASSED  ] 108 tests.
```

So the test has exactly the teeth the commit message claims — on one of the two sides. The other half of the hot path is unguarded, and it is the half that matters most: `docs/design-r5-gap-visibility.md:280-282` notes the pub side is the genuinely multi-threaded one, but recv is where rclcpp delivers every inter-process message.

### What this costs if it regresses

This PR's own bench prices the accumulator at **+25 CPU ns/msg** (measured below). `README`/`ROADMAP.md:152` promise "off by default costs nothing"; `topic_registry.hpp:116-117` promises the same. A refactor that drops either receive gate silently makes jitter mandatory on the receive path, and the fast lane, the TSan lane, and this test all stay green.

### Research

This is the standard vacuous-assertion failure mode: an assertion whose subject cannot vary under the code path being tested. Mutation-testing literature (Jia & Harman, *An Analysis and Survey of the Development of Mutation Testing*, IEEE TSE 2011) calls these *equivalent-passing* assertions — they raise the appearance of coverage while the mutant survives. The PR body's mutation discipline is exactly right; it just wasn't applied to the recv leg. Note `GapTracking.PubAndRecvTrackedSeparately` (`:263`, pre-existing) has the same shape — it asserts `EXPECT_FALSE(s->has_recv_max_dt)` with no subscription registered either, which is why the gap went unnoticed.

### Recommendation

Drive an actual delivery in the disabled phase, mirroring the publish half:

```cpp
TEST(GapTracking, DisabledPathNeverEntersTheAccumulator) {
    TopicRegistry reg;
    ASSERT_FALSE(reg.gapTracking());
    const void* pub = reinterpret_cast<const void*>(0x1010);
    const void* sub = reinterpret_cast<const void*>(0x1011);
    const void* rcl = reinterpret_cast<const void*>(0x1012);
    const void* cb  = reinterpret_cast<const void*>(0x1013);
    reg.onNodeInit(kNode, "talker", "/");
    reg.onPublisherInit(pub, kNode, "/chatter");
    reg.onSubscriptionInit(sub, kNode, "/chatter");
    reg.onRclcppSubscriptionInit(rcl, sub);
    reg.onCallbackAdded(cb, rcl);

    // Both sides, twice each: the first call resolves through the shared-lock path, the
    // second hits the TLS slot — so BOTH gap call sites per side run while disabled.
    reg.onPublish(pub);
    reg.onCallbackStart(cb, /*intra=*/false);
    std::this_thread::sleep_for(std::chrono::milliseconds(5));
    reg.onPublish(pub);
    reg.onCallbackStart(cb, /*intra=*/false);

    reg.setGapTracking(true);
    auto stats = reg.snapshot(1.0, /*fold_open_gap=*/true);
    const auto* s = findStat(stats, "/chatter");
    ASSERT_NE(s, nullptr);
    EXPECT_EQ(s->pub_inter_count, 2u);
    EXPECT_EQ(s->recv_inter_count, 2u);   // counting stayed live on BOTH sides
    EXPECT_FALSE(s->has_pub_max_dt)  << "disabled publish path left a timestamp";
    EXPECT_FALSE(s->has_recv_max_dt) << "disabled receive path left a timestamp";
}
```

Two `onCallbackStart` calls are needed, not one: the first misses the TLS slot and takes the resolve path (`:360`), the second hits (`:320`). Re-run the two-gate mutation afterwards — it should go red.

test/stress/test_concurrency.cpp
```diff
@@ -101,6 +102,76 @@ TEST(ConcurrencyStress, ManyThreadsManyTopicsExactTotals) {
...
+//   - max_dt_ms can never exceed the measured wall time of the whole test — the underflow value
+//     is ~570 years, so one unguarded reordered pair fails this bound.
+TEST(ConcurrencyStress, JitterMaxIsRaceFree) {
+    const int kThreads = envInt("ROS2_PULSE_STRESS_THREADS", 8);
+    const int kRounds = envInt("ROS2_PULSE_STRESS_ROUNDS", 1500);
+    const int kStallMs = envInt("ROS2_PULSE_STRESS_STALL_MS", 25);
+
+    TopicRegistry reg;
+    reg.setGapTracking(true);
+    reg.onPublisherInit(pubHandle(0), nullptr, "/pub_0");
```
TanayK07, line 122: ## 🔬 ISSUE — every handle in this file hashes to TLS slot 0, so the race test never runs the production fast path

**Severity:** HIGH — 2 of every 3 hot-path calls miss the thread-local cache and take the shared lock. `onPublish` misses **100%** of the time, so `publishCount`'s TLS-hit `noteArrival` call (`topic_registry.cpp:288-290`) gets zero concurrent coverage — and the shared-lock detour inserts a globally contended atomic in front of every `noteArrival`, which perturbs the exact reordering window this test exists to exercise.

### The problem

The handle helpers (`test_concurrency.cpp:50-53`) space topics by 1:

```cpp
const void* pubHandle(int i) { return H(0x100000u + static_cast<uintptr_t>(i)); }
const void* subHandle(int i) { return H(0x200000u + static_cast<uintptr_t>(i)); }
const void* rclSub(int i)    { return H(0x300000u + static_cast<uintptr_t>(i)); }
const void* cbHandle(int i)  { return H(0x400000u + static_cast<uintptr_t>(i)); }
```

`tlsSlotFor` (`topic_registry.cpp:48-53`) indexes with `((p >> 4) ^ (p >> 10)) & 255`. The `>> 4` deliberately discards the low four bits — the comment says so: *"Live-object addresses have zeroed alignment bits; >>4 drops them"*. These synthetic handles have **meaning** in those bits, so `i` is thrown away for every `i < 16`:

| handle | `p>>4` | `p>>10` | xor | `& 255` |
|---|---|---|---|---|
| `pubHandle(0..5)` `0x100000+i` | `0x10000` | `0x400` | `0x10400` | **0** |
| `subHandle(0..5)` `0x200000+i` | `0x20000` | `0x800` | `0x20800` | **0** |
| `rclSub(0..5)`  `0x300000+i` | `0x30000` | `0xC00` | `0x30C00` | **0** |
| `cbHandle(0..5)` `0x400000+i` | `0x40000` | `0x1000` | `0x41000` | **0** |

All four families, all six topics: slot 0.

### Measured, not argued

Replaying this test's exact call sequence (8 threads x 1500 rounds x 2 hammers) against the real `TopicRegistry` and reading the `sharedLockLookups()` observability hook the header advertises at `topic_registry.hpp:159-164`:

```
PR handles       ops=72000  shared_lock_lookups=48000 (66.7% of ops)  write_lock=1
spread handles   ops=72000  shared_lock_lookups=32    ( 0.0% of ops)  write_lock=1
```

66.7% is exactly 2 per round, and the arithmetic identifies which two. Per round the slot churns `cb -> pub -> cb`:

1. `onPublish(pubHandle(0))` — slot holds `cbHandle(0)` from the previous round ⇒ **miss**, always.
2. `onCallbackStart(cb, intra=true)` — slot holds `pubHandle(0)` ⇒ **miss**.
3. `onCallbackStart(cb, intra=false)` — slot holds `cb` ⇒ hit.

So the publish-side TLS-hit branch is taken **zero** times in the whole test, and the recv-side one is taken on 1 op in 3.

### Why this weakens the race it is written to catch

`noteArrival` samples `now` and then exchanges; the window for a reordered pair is the few nanoseconds between those two instructions. On the miss path, every call first executes `m_shared_lock_lookups.fetch_add(1, relaxed)` (`topic_registry.cpp:294`, `:333`) — a single process-wide cache line that all 8 threads RMW before reaching `noteArrival`. That serializing bounce, plus the `shared_lock` acquire and the `unordered_map` probe, spreads the threads apart and makes the interleaving far less dense than production, where the same traffic is a bare `fetch_add` + clock read with nothing shared in between. The mutation still fails 3/3 as the commit message reports, so the test is not toothless — but it samples a materially easier race than the code ships with, and the branch it never runs at all is the one that executes on essentially every real message.

### Research

This is the same class of defect the repo already documented and fixed once: `docs/KNOWN_ISSUES.md:262` records that the old bench used *"a single fixed key (`bench/hotpath_bench.cpp`: `t % kTopics` constant per thread) -> 100% cache hit"* and therefore measured a pattern the robot never runs. This is that bug's mirror image — a synthetic key distribution producing 0% cache hit — and it lives in the test that is supposed to certify the cache-and-gap interaction. `issue-15-tls-cache-stride-thrash.md` is the whole reason the index gained the `^ (p >> 10)` fold; a test whose keys collapse to one slot cannot regress any of it.

### Recommendation

Space handles by 16 so `>> 4` preserves the index, and offset the four families so they never share a slot:

```cpp
// tlsSlotFor() indexes on ((p>>4) ^ (p>>10)) & 255 and DISCARDS the low four bits, so
// consecutive synthetic handles would all land in slot 0 and every call would fall to the
// shared lock. Space by 16, and give each family its own 64-slot band, so the cache is
// exercised the way production exercises it (verified: distinct slots for i < 64).
const void* pubHandle(int i) { return H(0x100000u + (static_cast<uintptr_t>(i +   0) << 4)); }
const void* subHandle(int i) { return H(0x100000u + (static_cast<uintptr_t>(i +  64) << 4)); }
const void* rclSub(int i)    { return H(0x100000u + (static_cast<uintptr_t>(i + 128) << 4)); }
const void* cbHandle(int i)  { return H(0x100000u + (static_cast<uintptr_t>(i + 192) << 4)); }
```

Verified all-distinct for 4 families x 16 topics, and it drops the miss rate from 66.7% to 0.04% (32 of 72000 — just the per-thread cold start). Worth pairing with an explicit assertion so the property cannot silently rot:

```cpp
// Steady state must be served from the TLS cache; ~one miss per thread per endpoint is the
// cold-start floor. A regression here means the test is no longer covering the fast path.
EXPECT_LT(reg.sharedLockLookups(), static_cast<uint64_t>(kThreads) * 8);
```

`ManyThreadsManyTopicsExactTotals` shares these helpers and has the same blind spot, so the fix covers both.

bench/hotpath_bench.cpp
```diff
@@ -79,6 +111,40 @@ int main(int argc,char** argv){ int th=argc>1?atoi(argv[1]):8; long tot=(long)th
...
+    double sN=run(th,[&](int){ volatile uint64_t x=0; for(long i=0;i<kPer;i++) x+=Clock::now().time_since_epoch().count(); });
+    clk_n=sN*1e9/kPer;  // per-thread = CPU ns/op (threads run the whole wall interval)
+    std::printf("%-28s %7.1f CPU ns/op\n","raw steady_clock (th thr)", clk_n); }
+
+  // The published R5 figure is PER-MESSAGE CPU cost. The leg convention above is wall time
+  // amortized over all threads' ops, which divides the per-message cost by the parallelism —
+  // honest for throughput, misleading for "what does one message pay". Every thread runs the
+  // whole interval, so CPU ns/msg = amortized ns/op x threads. Print both so neither reading
+  // is left as an exercise.
+  const double d_fixed=(gap_fixed-base_fixed)*th, d_alt=(gap_alt-base_alt)*th;
```
TanayK07, line 144: ## 🔬 ISSUE — the delta is a single unpaired fixed-order run with no SEM, which is the methodology this repo's own docs forbid

**Severity:** HIGH — the figure it is meant to reproduce is quoted as `+24.366 +/- 0.023 ns` (SEM). Measured here, this harness's `alt-4` delta swings **27.2 to 31.2 CPU ns/msg across single runs (+/-14%)** and the attribution percentage swings **90% to 96%**. A one-shot number cannot confirm or refute the published one.

### The problem

`base_fixed`, `base_alt`, `gap_fixed`, `gap_alt` are each measured exactly once, always in that order, and `d_fixed`/`d_alt` are the raw differences scaled by `th`. There is no repetition, no order alternation, and no dispersion estimate — yet the output is printed as *the* R5 cost figure.

### Theory / proof — the repo already wrote this rule down

`bench/run_overhead_repeated.sh:1-7`, the script that invokes this very binary:

> `# Rigorous end-to-end overhead: PAIRED trials (baseline vs ours back-to-back) with the arm order`
> `# ALTERNATED each trial, so slow machine drift and any warm-up/order effect cancel in the`
> `# per-trial difference. Reports every sample plus mean/SD/SEM of the paired diffs — a delta is`
> `# only believable if it clears ~2x the SEM (the KNOWN_ISSUES #15 hunt showed single runs swing`
> `# +/-4% and a fixed arm order can masquerade as overhead).`

`docs/design-r5-gap-visibility.md:165` states the method used to produce the number being reproduced: *"Paired order-alternated, N=15, 8 threads"*. And its own test plan, line 529, specifies the requirement for this exact change:

> `bench/hotpath_bench.cpp` gains a jitter leg, **measured with the paired order-alternated harness**.

The jitter leg landed; the harness discipline did not. The fixed arm order is precisely the failure mode the shell script names: `base` always runs first, into a colder cache and a not-yet-ramped clock domain, and `gap` always second.

### Measured on this branch at `cd67b45` (16-core box, `g++ -O2 -std=c++17 -pthread`)

Five consecutive runs of `./hb 8`:

| run | `fixed` delta | `alt-4` delta | clock attribution |
|---|---|---|---|
| 1 | +25.1 | +27.7 | 92% |
| 2 | +25.7 | +28.3 | 90% |
| 3 | +24.8 | +28.3 | **96%** |
| 4 | +25.7 | +27.6 | 93% |
| 5 | +25.7 | +27.4 | 91% |

Six more interleaved runs put `alt-4` between +27.2 and +31.2. The attribution line is the worst case: it is a ratio of two independently noisy quantities, so its variance exceeds either input's. The header comment at line 9 says the attribution is printed *"so it stays checkable rather than folkloric"* — but any single run can be quoted as 90% or as 96%, and the PR body quotes 93% against a published 96%. Nothing distinguishes them without an error bar.

### What breaks

Nothing today; the numbers land near the published figure. It breaks the first time someone runs this on a busier box, a different core count, or after a hot-path change, reads one line of output, and either (a) "confirms" the +24 ns figure that has actually drifted, or (b) files a regression against a 14% sampling swing. Since `run_overhead_repeated.sh:59` runs this with no repetition (`/work/hb 8`, once), that is the default experience.

### Research

Paired, order-randomised measurement with reported dispersion is the standard for microbenchmarks whose effect size sits near the noise floor. Chen & Revels, *Robust Benchmarking in Noisy Environments* (arXiv:1608.04295), and Google Benchmark's `--benchmark_repetitions` + `_stddev`/`_cv` aggregates both exist for this: a single timing sample of a nanosecond-scale operation on a shared host has no defensible confidence interval. Curtsinger & Berger, *STABILIZER* (ASPLOS 2013), demonstrates that fixed layout and fixed execution order alone shift measured deltas by more than the effects usually being reported — which is exactly why the arm order must alternate, not merely repeat.

### Recommendation

Repeat the pair inside the binary, alternate the order, and print mean +/- SEM of the paired diffs — the same shape `run_overhead_repeated.sh` already implements in awk:

```cpp
// Paired + order-alternated, matching bench/run_overhead_repeated.sh: the arms run back to back
// and swap order each trial, so drift and warm-up cancel in the per-trial DIFFERENCE. A single
// unpaired run of these legs swings +/-14% on alt-4 — far wider than the effect's own precision.
const int reps = argc > 2 ? atoi(argv[2]) : 5;
std::vector<double> dfix, dalt;
for (int i = 0; i < reps; ++i) {
  double bf, ba, gf, ga;
  auto plain = [&]{ New256<false> n; for(int j=0;j<kTopics;j++) n.reg(ks[j]);
                    bf = leg("cache-256 TLS (fixed)", n, th, tot, ks, false);
                    ba = leg("cache-256 TLS (alt-4)", n, th, tot, ks, true); };
  auto withg = [&]{ New256<true>  n; for(int j=0;j<kTopics;j++) n.reg(ks[j]);
                    gf = leg("cache-256+gap (fixed)", n, th, tot, ks, false);
                    ga = leg("cache-256+gap (alt-4)", n, th, tot, ks, true); };
  if (i % 2 == 0) { plain(); withg(); } else { withg(); plain(); }
  dfix.push_back((gf - bf) * th);
  dalt.push_back((ga - ba) * th);
}

auto report = [&](const char* label, std::vector<double>& d){
  double m = 0; for (double x : d) m += x; m /= d.size();
  double v = 0; for (double x : d) v += (x - m) * (x - m);
  const double sd = d.size() > 1 ? std::sqrt(v / (d.size() - 1)) : 0.0;
  std::printf("  %-6s: %+6.2f +/- %.2f CPU ns/msg (SEM, N=%zu)%s\n",
              label, m, sd / std::sqrt((double)d.size()), d.size(),
              m > 2 * sd / std::sqrt((double)d.size()) ? "" : "   [NOT RESOLVED — below 2x SEM]");
};
report("fixed", dfix);
report("alt-4", dalt);
```

Then update `bench/RESULTS.md`'s per-operation table (currently lines 60-75, still pre-gap) with the `cache-256+gap` rows and their SEMs, so the `+24.4 ns` in `docs/ROADMAP.md:152` points at a committed measurement rather than at a harness that has to be re-run by hand to mean anything.

bench/hotpath_bench.cpp
```diff
@@ -46,12 +52,37 @@ struct NewMap : RegBase {
     std::shared_lock<std::shared_mutex> l(m); auto it=r.find(k);
     if(it!=r.end()){ it->second->c.fetch_add(1,std::memory_order_relaxed); s={k,it->second}; } } };
 
+// current design: 256-slot TLS cache + stride-breaking hash (mirrors src/core/topic_registry.cpp
+// after KNOWN_ISSUES #15). Optionally runs the R5 gap accumulator on every hit — an exact replica
+// of noteArrival: exchange partitions the predecessor timestamp between racing threads, and the
+// now > prev guard drops reordered pairs instead of letting the unsigned subtraction underflow.
+struct GapCtr { std::atomic<uint64_t> c{0}; std::atomic<uint64_t> last{0}; std::atomic<uint64_t> maxdt{0}; };
```
TanayK07, line 59: ## 🔬 GAP ANALYSIS — `GapCtr` is 24 bytes; the counter it models is 104, and the difference inflates the reported `alt-4` cost by +2.7 CPU ns/msg

**Severity:** MEDIUM — the leg is documented as *"an exact replica of `noteArrival`"* and the comment above claims it *"mirrors `src/core/topic_registry.cpp`"*. The instruction sequence is an exact replica; the memory layout is not, and for this measurement the layout is most of the physics.

### The problem

```cpp
struct GapCtr { std::atomic<uint64_t> c{0}; std::atomic<uint64_t> last{0}; std::atomic<uint64_t> maxdt{0}; };
```

24 bytes, three hot fields at offsets 0/8/16 — always one cache line, and small enough that `operator new` packs several endpoints into the same line. The real `sTopicCounter` (`include/ros2_pulse/core/topic_registry.hpp:23-47`) leads with a `std::string`, a `bool`, and four count buckets before the gap fields.

Measured against the real header on this branch:

```
sizeof(sTopicCounter)=104   pub_inter=40  recv_inter=56  pub_last_ns=72  recv_last_ns=80
sizeof(GapCtr)=24           c=0           last=8        maxdt=16
```

The two fields the gap path touches per message are **32 bytes apart** in production and **8 bytes apart** in the model. Allocating 128 counters exactly as `counterForTopic` does (`topic_registry.cpp:146`):

```
pub_inter and pub_last_ns on DIFFERENT cache lines: 64/128  (same line: 64/128)
```

That 64/128 is not a coincidence — `docs/design-r5-gap-visibility.md:348` records the identical figure from the design-time analysis: *"R5 inline (96 B): straddling 64/128 | jitter fields on a different line from counts 64/128"*. The design doc measured this property and the harness that is supposed to reproduce the design doc's number models the other case.

### Measured effect — paired, order-alternated, N=6

Rebuilt this bench with `GapCtr` mirroring `sTopicCounter`'s real field layout (string + bool + four count buckets + four gap fields, 104 B) and nothing else changed:

| trial | order | PR `GapCtr` (24 B) | layout-faithful (104 B) | diff |
|---|---|---|---|---|
| 1 | 24B,104B | +28.4 | +26.6 | +1.8 |
| 2 | 104B,24B | +27.2 | +25.5 | +1.7 |
| 3 | 24B,104B | +29.9 | +27.5 | +2.4 |
| 4 | 104B,24B | +30.4 | +27.0 | +3.4 |
| 5 | 24B,104B | +27.8 | +25.8 | +2.0 |
| 6 | 104B,24B | +31.2 | +26.2 | +5.0 |

`alt-4` CPU ns/msg. **Paired diff = +2.72 +/- 0.52 (SEM, N=6)** — 5.2x the SEM, so it clears the repo's own believability bar. The `fixed` leg is unaffected (24.8-25.7 either way).

The direction is over-statement, and the mechanism is inter-endpoint false sharing, not intra-struct straddling: at 24 bytes the four endpoints a thread alternates over pack into one or two lines that all 8 threads then fight over. The real allocator does not do this — `design-r5-gap-visibility.md:331-333` measured *"the observed `make_unique` stride is a uniform 80 bytes, which exceeds the 64-byte line, so consecutive counters' hot fields land on distinct lines anyway (measured: 8 adjacent endpoints -> 8 distinct lines)"*. The model reintroduces exactly the false sharing the real layout was measured not to have.

Today the `fixed` leg is close enough that the headline reproduction reads fine. What breaks is the `alt-4` figure — the one `docs/KNOWN_ISSUES.md:275` and `bench/RESULTS.md` designate as the number that matters in practice (*"The alternating-pattern number is the one that matters"*). It is currently ~10% high for a structural reason, and it will move unpredictably the moment anyone changes `sTopicCounter`'s field order, because the model does not track it.

### Research

Two distinct effects, both well characterised: intra-object line straddling and inter-object false sharing. Intel's *Avoiding and Identifying False Sharing Among Threads* and Bolosky & Scott (*False Sharing and its Effect on Shared Memory Performance*, SEDMS IV) establish that shared-line RMW traffic dominates nanosecond-scale per-operation costs; Chandra et al.'s cache-contention work shows an under-sized synthetic object is the classic way to manufacture it accidentally. This repo has already litigated the question once — the `alignas(64)` section at `design-r5-gap-visibility.md:311-377` is retained as a *negative* result specifically so nobody redoes the analysis, and it concludes R5 needs no layout change *given the real 80-byte allocation stride*. A model that discards that stride is measuring the hypothesis the design doc rejected.

### Recommendation

Mirror the real struct's layout so the model tracks the header if it changes:

```cpp
// Layout-faithful to sTopicCounter (include/ros2_pulse/core/topic_registry.hpp): 104 B, with the
// count buckets 32 B ahead of the gap fields, so the +gap leg pays the same second-line touch
// real endpoints pay (measured 64/128) and the alternating leg sees the same ~80 B allocation
// stride instead of manufacturing false sharing a 24 B struct would. Field ORDER matters here,
// not just size: shrinking this to the three hot fields over-states the alt-4 delta by
// +2.7 +/- 0.5 CPU ns/msg (paired, N=6).
struct GapCtr {
  std::string topic;                                              // cold, but it is what pushes
  bool seen{false};                                               // the atomics to offset 40
  std::atomic<uint64_t> c{0}, pub_intra{0}, recv_inter{0}, recv_intra{0};
  std::atomic<uint64_t> last{0}, recv_last{0}, maxdt{0}, recv_max{0};
};
static_assert(sizeof(GapCtr) == 104, "GapCtr must track sTopicCounter's layout");
```

Alternatively `#include "ros2_pulse/core/topic_registry.hpp"` and use `sTopicCounter` directly — the bench is standalone `g++` today, but the header is header-only for these fields and that removes the drift risk entirely.

bench/hotpath_bench.cpp
```diff
@@ -79,6 +111,40 @@ int main(int argc,char** argv){ int th=argc>1?atoi(argv[1]):8; long tot=(long)th
...
+  // amortized over all threads' ops, which divides the per-message cost by the parallelism —
+  // honest for throughput, misleading for "what does one message pay". Every thread runs the
+  // whole interval, so CPU ns/msg = amortized ns/op x threads. Print both so neither reading
+  // is left as an exercise.
+  const double d_fixed=(gap_fixed-base_fixed)*th, d_alt=(gap_alt-base_alt)*th;
+  std::printf("\nR5 gap cost, +gap minus plain (ROS_TOPIC_STATS_JITTER=1):\n");
+  std::printf("  fixed : %+6.1f CPU ns/msg  (wall-amortized %+5.1f)\n", d_fixed, (gap_fixed-base_fixed));
+  std::printf("  alt-4 : %+6.1f CPU ns/msg  (wall-amortized %+5.1f)\n", d_alt, (gap_alt-base_alt));
+  std::printf("  clock read alone: %.1f CPU ns -> %.0f%% of the fixed-leg delta\n",
+              clk_n, 100.0*clk_n/(d_fixed>0?d_fixed:1));
```
TanayK07, line 149: ## 🔬 ISSUE — the divide guard turns a failed measurement into a confident-looking percentage

**Severity:** LOW — no wrong behaviour in the happy path, but the one output line whose stated job is to keep the attribution *"checkable rather than folkloric"* prints a fabricated number when the measurement fails.

### The problem

```cpp
std::printf("  clock read alone: %.1f CPU ns -> %.0f%% of the fixed-leg delta\n",
            clk_n, 100.0*clk_n/(d_fixed>0?d_fixed:1));
```

The `d_fixed > 0 ? d_fixed : 1` guard prevents a division by zero and then divides by **1 nanosecond**, emitting a percentage that reads as a real result. With a typical `clk_n` of 23 CPU ns, a non-positive `d_fixed` prints:

```
  clock read alone: 23.2 CPU ns -> 2320% of the fixed-leg delta
```

Nothing marks that line as invalid. A reader scanning the output sees a number with a `%` after it.

### When `d_fixed <= 0` actually happens

Not hypothetical for a delta of two independently-timed legs at this scale:

- a co-tenant process lands on the base leg (it always runs first — see the pairing comment on line 144),
- `th` exceeds available cores, so wall time is dominated by scheduling rather than per-op cost,
- someone builds at `-O0` or with a sanitizer, where both legs are dominated by instrumentation,
- a future run on a platform where `steady_clock` is not vDSO-backed, which `design-r5-gap-visibility.md:531-533` explicitly flags as the Orin case worth re-checking.

On any of those the harness reports a garbage attribution instead of reporting that it could not attribute.

### Theory

A sentinel that silently substitutes a plausible-looking value is the anti-pattern the rest of this codebase avoids by construction. `takeGap` (`topic_registry.cpp:98-105`) returns `false` rather than reporting `0.000` for exactly this reason, with the rationale spelled out: *"Reporting 0.000 would read as 'perfectly regular', which is the best possible value, inferred from a single data point."* `pulse-check` reserves **exit 2** for "cannot judge" instead of collapsing it into 0 or 1 (`design-r5-gap-visibility.md:149-151`). The bench should follow the same rule its own subject follows: absence of a measurement is not a measurement.

### Recommendation

```cpp
if (d_fixed > 0.0) {
  std::printf("  clock read alone: %.1f CPU ns -> %.0f%% of the fixed-leg delta\n",
              clk_n, 100.0 * clk_n / d_fixed);
} else {
  // Same discipline as takeGap()/pulse-check exit 2: no measurement is not a measurement of
  // zero. A non-positive delta means the run was too noisy to attribute, not that the clock
  // read is free — say so instead of dividing by a sentinel.
  std::printf("  clock read alone: %.1f CPU ns -> attribution N/A (fixed-leg delta was %+.1f;"
              " re-run on a quiet box)\n", clk_n, d_fixed);
}
```

test/unit/test_gap_tracking.cpp
```diff
@@ -203,6 +203,38 @@ TEST(GapTracking, OpenGapFoldedSoSilentEndpointGrows) {
 // ...but the exit window must NOT fold. rclcpp teardown stops traffic before the process exits,
 // so the open interval there measures the shutdown sequence and would fire every gap rule on a
 // perfectly healthy stop.
+// The R5 design's "DisabledReadsNoClock": default-off must cost zero, i.e. the disabled hot
```
TanayK07, line 206: ## 🔬 ISSUE — this insertion orphans the comment block above it onto the wrong test

**Severity:** LOW — no behaviour change, but two tests in a heavily-commented file now carry misattributed documentation, and one of them is the exit-window suppression case.

### The problem

Lines 203-205 were written as the preamble to `OpenGapNotFoldedWhenExiting`, continuing the `...but` clause from `OpenGapFoldedSoSilentEndpointGrows` directly above it:

```cpp
// ...but the exit window must NOT fold. rclcpp teardown stops traffic before the process exits,
// so the open interval there measures the shutdown sequence and would fire every gap rule on a
// perfectly healthy stop.
// The R5 design's "DisabledReadsNoClock": default-off must cost zero, ...   <-- new block starts
TEST(GapTracking, DisabledPathNeverEntersTheAccumulator) {
```

The new test landed between that comment and the `TEST` it describes. As of this commit:

- `DisabledPathNeverEntersTheAccumulator` is introduced by three lines about rclcpp teardown and the exit window, which it has nothing to do with — it never calls `snapshot(..., false)` and never simulates shutdown.
- `OpenGapNotFoldedWhenExiting` (line 238) is now the only gap test in the file with no comment at all, having lost the rationale for the one behaviour that is counter-intuitive enough to need it.

### Why it matters here specifically

The lost text is not decorative. `docs/design-r5-gap-visibility.md:109-122` devotes a whole section to it (*"…and it must be suppressed in the exit window"*), because the fold firing at `atexit` would make `max_gap_ms` alert on every clean shutdown, deterministically. The `...but` is a contrast against the test immediately above, and the contrast is now split across an unrelated test. This file's convention is that every `TEST` carries the reasoning that makes its assertions non-arbitrary — the misplacement is the one thing that convention cannot survive silently, since nothing checks it.

### Recommendation

Move the new test below `OpenGapNotFoldedWhenExiting` (its natural neighbour is `PubAndRecvTrackedSeparately`, which is also a gate/independence test), or keep the position and restore the orphan:

```cpp
// A fully-silent endpoint ... (OpenGapFoldedSoSilentEndpointGrows above)

// The R5 design's "DisabledReadsNoClock": default-off must cost zero ...
TEST(GapTracking, DisabledPathNeverEntersTheAccumulator) { ... }

// ...but the exit window must NOT fold. rclcpp teardown stops traffic before the process exits,
// so the open interval there measures the shutdown sequence and would fire every gap rule on a
// perfectly healthy stop.
TEST(GapTracking, OpenGapNotFoldedWhenExiting) { ... }
```

If it stays where it is, reword the leading `...but` — it now reads as a continuation across an intervening test.

bench/hotpath_bench.cpp
```diff
@@ -79,6 +111,40 @@ int main(int argc,char** argv){ int th=argc>1?atoi(argv[1]):8; long tot=(long)th
     leg("single-entry TLS (alt-4)",   n, th, tot, ks, true); }
 
   { NewMap n; for(int i=0;i<kTopics;i++) n.reg(ks[i]);
-    leg("mini-map TLS (fixed)",       n, th, tot, ks, false);
-    leg("mini-map TLS (alt-4)",       n, th, tot, ks, true); }
+    leg("mini-map-16 TLS (fixed)",    n, th, tot, ks, false);
+    leg("mini-map-16 TLS (alt-4)",    n, th, tot, ks, true); }
+
+  // Current design without and with the R5 gap accumulator. The pairwise delta of the same leg
```
TanayK07, line 117: ## 🔬 GAP ANALYSIS — the harness is committed but still unbuilt by CI, and `RESULTS.md` was not updated, so the `+24 ns` figure still has no regression gate

**Severity:** LOW — the PR closes the "no committed harness" half of the house rule. The other half, *published numbers stay checkable*, is not yet closed.

### The problem

The commit message frames the goal as the house rule that *"every published number has a harness"*. After this change the harness exists and is reachable, but nothing runs it where a drift would be noticed:

| Path | State after this PR |
|---|---|
| `bench/CMakeLists.txt` | builds `stress_nodes` only — `hotpath_bench.cpp` is not a target |
| `.github/workflows/ci.yml` | no bench lane; the fast and sanitizer lanes build `ROS2_PULSE_STANDALONE` targets only |
| `bench/run_overhead_repeated.sh:58-59` | does compile and run it — but requires a privileged container, a mounted `/pkg`, and a manual invocation |
| `bench/README.md` "Files" section | lists `stress_nodes.cpp`, `bpftrace_probe.bt`, `run_bakeoff.sh`, `RESULTS.md` — not `hotpath_bench.cpp` |
| `bench/RESULTS.md:60-75` | per-operation table still ends at *"current TLS cache (post-#15), alt-4"*; no `cache-256+gap` rows, no gap delta, no clock-attribution row |
| `docs/ROADMAP.md:152` | still asserts **+24.4 +/- 0.02 ns/msg (96% of it the vDSO clock read)** with no committed measurement backing it |

So the number a reader encounters (`ROADMAP.md:152`) and the number this harness produces live in different places, and only the first is published.

Today: correct, and someone who reads the bench source can reproduce it by hand. What breaks is the next hot-path change — an extra atomic in `noteArrival`, a `sTopicCounter` field reorder, a different clock. The delta moves, every lane stays green, and `ROADMAP.md` keeps asserting `+24.4` because no artifact contradicts it. That is the same failure mode this PR was written to fix, one level up: `hotpath_bench.cpp` previously modelled the pre-#15 16-slot cache as "current" for however long it took someone to notice, precisely because nothing exercised it.

### Research

This is the "committed but unwired" state that benchmark-regression tooling exists to escape. Google Benchmark ships `tools/compare.py` with a U-test against a stored baseline; Rust's `criterion` persists baselines under `target/criterion` and fails on a configured regression threshold; `pytest-benchmark` has `--benchmark-compare-fail`. The common design is a stored artifact plus a comparison step, not a harness a human must remember to run. Given a nanosecond-scale delta on shared CI runners, a hard gate would be too flaky — but a recorded artifact with an error bar is not, and that is what `RESULTS.md` already is for the other legs.

### Recommendation

Cheapest complete version, no CI gate required:

1. Add the target so it at least compiles in every build (a bench that stops compiling is the most common way this rots):
   ```cmake
   # bench/CMakeLists.txt — pure C++, no ROS deps; built so a hot-path change that breaks the
   # harness fails loudly instead of silently going unmeasured.
   add_executable(hotpath_bench hotpath_bench.cpp)
   target_compile_features(hotpath_bench PRIVATE cxx_std_17)
   target_link_libraries(hotpath_bench Threads::Threads)
   ```
2. Append the new legs to `bench/RESULTS.md`'s per-operation table with the paired SEMs from the pairing change (see the comment on line 144), and add the run date + host, matching the `2026-08-01 run of hotpath_bench.cpp inside each distro container` convention already used there.
3. Add `hotpath_bench.cpp` to `bench/README.md`'s Files list with a one-line description of what the `+gap` leg measures.
4. Make `docs/ROADMAP.md:152` cite `bench/RESULTS.md` for the `+24.4` figure, so the published number and the measurement are one hop apart.

PR #27, opened 2026-08-08
feat: ROS_TOPIC_STATS_QUIET=1 silences the banner: never diagnostics (ROADMAP R6)

Tanay10x:
## What

ROADMAP R6: `ROS_TOPIC_STATS_QUIET=1` suppresses the `[ros2_pulse] active` stderr banner for deployments that parse the wrapped process's stderr and can't tolerate a foreign line.

**Deliberately narrow scope:** error diagnostics always print. Every diagnostic QUIET could have eaten (unreadable/malformed/zero-rule spec, unwritable output file) reports the disabling of something the operator explicitly configured, and *silently disabled alerting* is a bug class 0.2.0 fixed (directory-as-spec parsing to a zero-rule no-op, invisible UTF-8 BOM failures). A QUIET that swallowed those would reintroduce that class as a supported flag. Rationale at the suppression site, in the README row, and pinned by test.

Bonus semantics pinned: only the exact documented `"1"` opts in (existing `envFlag()` contract), `QUIET=true` keeps the banner, so a half-set flag fails loud rather than half-silencing.

## Verification (TDD)

- Red: `integration_quiet` failed with the banner visibly leaking through `QUIET=1` before the feature commit; the two default-behavior tests passed, pinning the status quo.
- Green: 4/4 in `test_quiet.py`, then the full Humble integration suite 12/12 and standalone fast lane 104/104.
- New test launches the probed talker with `stderr=PIPE` (the existing harness discards stderr) and asserts: banner present by default, absent under QUIET, and a malformed `ROS_TOPIC_STATS_EXPECTED` still warns under QUIET.


src/probe/interposers.cpp
```diff
@@ -132,17 +132,30 @@ class ProbeRuntime {
...
+            // it exists for deployments that parse the wrapped process's stderr and can't
+            // tolerate a foreign line. Scope decided deliberately narrow: the one-shot error
+            // diagnostics (unreadable/malformed/zero-rule spec in loadRateSpec, unwritable
+            // output file in flush) are NOT gated on it, because each of those paths disables
+            // a feature the operator explicitly configured, and silently disabled alerting is
+            // a bug class 0.2.0 fixed (see CHANGELOG: a directory read as a spec parsed to a
+            // zero-rule no-op, a UTF-8 BOM failed a valid spec invisibly). A QUIET that also
+            // swallowed diagnostics would reintroduce that class as a supported flag.
+            // Pinned by test/integration/test_quiet.py.
+            if (!m_quiet) {
```
TanayK07, line 156: ## 🔬 GAP ANALYSIS: QUIET deletes the only signal for the *most likely* cause of silently-disabled alerting

**Severity:** HIGH — under `QUIET=1`, an operator whose `ROS_TOPIC_STATS_EXPECTED` was dropped by systemd/`sudo`/`ros2 launch` gets an empty stderr and a stats file indistinguishable from a healthy armed run. That is the failure class this PR's scope note says the flag must not reintroduce.

### The problem

The rationale block above (lines 136–145) argues the surviving diagnostics are sufficient:

> the one-shot error diagnostics (unreadable/malformed/zero-rule spec in loadRateSpec, unwritable output file in flush) are NOT gated on it, because each of those paths disables a feature the operator explicitly configured

That enumeration is complete for `loadRateSpec()`'s **error** branches — but it omits the branch that runs first:

```cpp
// interposers.cpp:76-79
const char* path = std::getenv("ROS_TOPIC_STATS_EXPECTED");
if (path == nullptr || *path == '\0') {
    return std::nullopt;          // <-- silent, deliberately
}
```

Unset returns `nullopt` with **no diagnostic**, and correctly so: warning here would fire in every unconfigured process on the robot. The banner was what covered that branch — and `ensureStarted()`'s own comment two lines up (132–134) says exactly this:

> every disabling path (**unset**, unreadable, malformed, zero-rule) lands on `spec=none`

So `spec=none` is load-bearing precisely for the one disabling path that has no stderr diagnostic, and line 146 removes it.

### Theory / proof

Alerting can be off for five reasons. Coverage under `QUIET=1` after this PR:

| Disabling path | stderr diagnostic | banner `spec=none` | visible with QUIET=1 |
|---|---|---|---|
| unreadable spec | yes (`cannot read …`) | yes | yes |
| malformed spec | yes (`invalid …`) | yes | yes |
| zero-rule spec | yes (`declares no topics…`) | yes | yes |
| **var unset / lost in transit** | **no** | yes | **no** |
| var never intended | no | yes | n/a (correct) |

Rows 4 and 5 are byte-identical from inside the process — `getenv` returns `nullptr` for both — so the probe can *never* warn on row 4 without spamming row 5. The banner never needed to tell them apart: it reported **state**, and the operator supplied the intent. Suppressing state reporting collapses row 4 into row 5 with no recovery path.

### Why row 4 is the ordinary case, not the exotic one

Env-var loss is the routine failure on exactly the deployments QUIET targets:

- `sudo` without `-E`: `env_reset` is the default in Debian/Ubuntu `/etc/sudoers`, and `env_keep` carries no `ROS_*`.
- systemd units start from an empty environment; a var arrives only via `Environment=`/`EnvironmentFile=`. A unit that sets `LD_PRELOAD` and `ROS_TOPIC_STATS_OUTPUT_FILE` but forgets `ROS_TOPIC_STATS_EXPECTED` still yields a running, logging, apparently-healthy probe.
- `ros2 launch` inherits the launching shell, but `Node(env=…)`, `SetEnvironmentVariable` scoping, and containerised composition (`component_container`) routinely reshape it.
- Docker `--env-file` typos; k8s ConfigMap key renames.

Contrast with the two bugs the scope note cites as the class to protect (CHANGELOG 0.2.0): a directory read as a spec, and a UTF-8 BOM. **Both require the operator to have successfully set the variable** and then hit a parsing edge. "The variable never arrived" is strictly more common than either.

Today (no QUIET): the probe prints `spec=none` at attach. The operator greps once, sees `spec=none` where they expected `spec=3 topics, 2 nodes`, and fixes the unit file.

With `QUIET=1`: stderr is empty (by design), and the stats file gives nothing back. `formatWindow()` (src/core/window_format.cpp:36-40) emits only `# ts_ns=… window_s=…` — there is no `spec=` field anywhere in the file — and `WARN` lines appear only on violation, so *absence of WARN* reads identically to "armed, all healthy". The bug surfaces the day a topic actually dies and no alert fires.

In fairness, one partial backstop exists: an operator who *also* runs `pulse-check --spec` from a watchdog re-derives verdicts from raw rates and would catch it. But `pulse-check` is optional and out-of-process, while the README row sells QUIET as safe on its own.

### Research

Instrumentation tools separate *destination* from *content*, and never let a verbosity switch remove state only the tool can report:

- glibc `ld.so`: `LD_DEBUG_OUTPUT` relocates diagnostics to a file; there is no flag that discards the "what got preloaded" state.
- Valgrind: `-q` suppresses non-error output, but `--log-file=`/`--log-fd=` exists so state is *moved*, not dropped.
- Linux `perf`: run configuration and event descriptors are written into `perf.data`'s header (`HEADER_CMDLINE`, `HEADER_EVENT_DESC`), so a session is self-describing regardless of terminal verbosity.
- `strace -o file`: the whole output leaves the tracee's fd 2 rather than being conditionally silenced.
- OpenTelemetry `Resource`: the run's configuration is stamped onto the stream itself, so a consumer never has to trust that a process was configured as intended.

Common shape: **put the configuration into the artifact you already write, not only onto the console you may be asked to silence.**

### Recommendation

Keep line 146 as written, and make the stats file self-describing so QUIET has nothing load-bearing left to take away:

```cpp
// ensureStarted(): compute spec_desc UNCONDITIONALLY, gate only the stderr write.
char spec_desc[64];
if (m_spec.has_value()) {
    std::snprintf(spec_desc, sizeof(spec_desc), "%zu topics, %zu nodes",
                  m_spec->topics.size(), m_spec->nodes.size());
} else {
    std::snprintf(spec_desc, sizeof(spec_desc), "none");
}
if (!m_quiet) {
    std::fprintf(stderr, "[ros2_pulse] active — interposing tracetools layer "
                         "(out=%s, period=%.1fs, spec=%s, jitter=%s)\n",
                 m_out_path.c_str(), m_period_s, spec_desc,
                 m_registry.gapTracking() ? "on" : "off");
}
// NEVER gated on QUIET: the stats file is not the wrapped process's stderr, so recording the
// probe's own configuration there costs a stderr-parsing deployment nothing — and it keeps
// spec=none recoverable for the ONE disabling path that has no stderr diagnostic (var unset).
m_header = std::string("# ros2_pulse spec=") + spec_desc +
           " jitter=" + (m_registry.gapTracking() ? "on" : "off") + "\n";
```

Emit `m_header` once as the first line of the first flushed block. It is a `#` comment, so `probe_harness.py`'s `_HEADER_RE` (anchored on `# ts_ns=`) ignores it by construction — worth confirming `src/core/log_reader.cpp` also tolerates a second `#` line before wiring it up.

That then lets `pulse-check` close the loop for free: log says `spec=none` but the user passed `--spec` → exit 2 with `probe ran with no spec armed — alerting was never on`. Same stance this project already took for `max_gap_ms` in the current Unreleased entry: *measurement absence is not a health verdict.*

Minimum viable alternative if that is out of scope for R6: state the blind spot in the README row and in this rationale block — "QUIET=1 also hides `spec=none`; confirm the spec armed once without QUIET, or gate on `pulse-check`" — so it lands as a documented trade-off rather than an unstated one.

README.md
```diff
@@ -109,6 +109,7 @@ A missing preload lib is non-fatal (`ld.so` warns and ignores), so it is safe to
...
 | `ROS_PULSE_EMIT_IDLE` | `0` | When `1`, also emit a `TOPIC /x 0.000000` line for a **declared-but-silent** topic (one with no traffic in the window). By default (`0`) such topics are omitted, so a large graph isn't padded with a zero line per silent topic every window. Only the publish-side `TOPIC` line is affected; `RECV` emits an explicit `0.000000` line for topics that have delivered at least once (so a **stalled** upstream stays visible) and omits never-active topics. |
 | `ROS_TOPIC_STATS_EXPECTED` | unset | Path to an expected-rate spec (below). When set, each window is checked at flush time and violations are appended as `WARN` lines. Unreadable or malformed specs warn once on stderr and disable alerting — never crash the host. |
+| `ROS_TOPIC_STATS_QUIET` | `0` | When `1`, suppress the `[ros2_pulse] active` stderr banner — for deployments that parse the wrapped process's stderr. Silences the **informational banner only**: error diagnostics (unreadable/malformed/zero-rule spec, unwritable output file) still print, because each of those disables something you explicitly configured, and silently disabled alerting is the 0.2.0 bug class this flag must not reintroduce. |
```
TanayK07, line 112: ## 🔬 GAP ANALYSIS: a boolean silence flag cannot satisfy the requirement this row states

**Severity:** HIGH — the documented use case is "deployments that parse the wrapped process's stderr", but `QUIET=1` leaves four stderr writes able to fire. Those deployments are still one malformed spec or one full disk away from the breakage the flag exists to prevent.

### The problem

The row promises a fix for stderr-parsing deployments and withdraws most of it in the same sentence:

> suppress the `[ros2_pulse] active` stderr banner — for deployments that parse the wrapped process's stderr. Silences the **informational banner only**: error diagnostics … still print

Both halves are individually correct. Together they describe a flag that does not close the stated requirement. A consumer that "can't tolerate a foreign line" tolerates zero foreign lines — not "zero except on the days something goes wrong", which are the days a parser crash costs the most.

Five writes can reach the wrapped process's stderr. QUIET gates one:

| Source | Gated by QUIET |
|---|---|
| `interposers.cpp:154` — `[ros2_pulse] active — …` | yes |
| `interposers.cpp:85` — `cannot read ROS_TOPIC_STATS_EXPECTED …` | no |
| `interposers.cpp:93` — `invalid ROS_TOPIC_STATS_EXPECTED …` | no |
| `interposers.cpp:102` — `… declares no topics and no nodes …` | no |
| `interposers.cpp:345` — `cannot open output file … — dropping windows` | no |

The last is the sharpest. It fires from the **flush thread**, so it can land at any window boundary — minutes into a run, long past any start-up sanity check the deployment performs — and it is reachable from ordinary operational events: disk fills, the output directory sits on a tmpfs that got recreated, an NFS mount drops.

### Theory — the two goals are not actually in tension

The PR frames this as a real dilemma: silence everything and reintroduce the 0.2.0 silently-disabled-alerting class, or keep diagnostics and accept a foreign line. That dilemma exists only because the flag controls **whether** output happens instead of **where** it goes. Moving the probe's own diagnostics off fd 2 satisfies both goals at once, unconditionally:

- the wrapped process's stderr becomes byte-for-byte what it would be without `LD_PRELOAD`, for **all five** writes, permanently;
- every diagnostic survives in full, at a destination the operator chose;
- no scope argument remains and no bug class can be reintroduced, because nothing is discarded.

### Research — destination, not mute, is the settled convention here

Every mature tool with this injection model exposes a destination:

| Tool | Mechanism | Note |
|---|---|---|
| glibc `ld.so` | `LD_DEBUG_OUTPUT=<prefix>` | writes `<prefix>.<pid>`, off the tracee's stderr |
| glibc mtrace | `MALLOC_TRACE=<file>` | env var names a file, not a bool |
| Valgrind | `--log-file=`, `--log-fd=N`, `--log-socket=` | `-q` exists, but as verbosity layered on top |
| LLVM sanitizers | `ASAN_OPTIONS=log_path=…` (appends `.pid`) | same problem, same answer |
| gperftools | `CPUPROFILE=`, `HEAPPROFILE=` | destination-only, no silence flag |
| `strace` / `ltrace` | `-o file`, `-o \|cmd` | canonical: get out of the tracee's fd 2 |

The pattern holds because all of these share this PR's constraint: they are injected into a process whose stdio contract belongs to somebody else. `LD_DEBUG_OUTPUT` is the closest analogue — same `ld.so` injection, same problem statement — and glibc chose relocation. There is no `LD_DEBUG_QUIET`.

### Recommendation

Add a destination variable and keep `QUIET` as the trivial special case, so this PR's tests and docs stay valid as-is.

```cpp
// interposers.cpp, near envFlag()
// Where the probe's OWN messages go. Default fd 2. An operator whose stderr is parsed sets
// ROS_TOPIC_STATS_LOG_FILE and gets every diagnostic AND a pristine wrapped stderr — the two
// goals QUIET can only trade against each other.
std::FILE* g_diag = stderr;   // never fclose'd: leaky by design, same as the runtime (#9)

void openDiagStream() {
    const char* p = std::getenv("ROS_TOPIC_STATS_LOG_FILE");
    if (p == nullptr || *p == '\0') return;
    // "ae" = append + O_CLOEXEC, so a fork+exec'd child does not inherit the fd. Apply the same
    // O_NOFOLLOW hardening the stats file got; on failure keep stderr and say so ONCE — a
    // diagnostic sink that fails silently is the exact bug this feature is about.
    if (std::FILE* f = std::fopen(p, "ae")) {
        std::setvbuf(f, nullptr, _IOLBF, 0);  // line-buffered: one-shot warns must not sit in a buffer
        g_diag = f;
    } else {
        std::fprintf(stderr, "[ros2_pulse] cannot open ROS_TOPIC_STATS_LOG_FILE '%s' (%s) — "
                             "diagnostics stay on stderr\n", p, std::strerror(errno));
    }
}
```

Then replace all five `std::fprintf(stderr, …)` sites in `interposers.cpp` with `g_diag`, call `openDiagStream()` first in the `ProbeRuntime` ctor (before `loadRateSpec()`, so spec diagnostics land in the file too), and document:

```markdown
| `ROS_TOPIC_STATS_LOG_FILE` | unset | Redirect **all** probe messages — banner and diagnostics — to this file instead of the wrapped process's stderr. Preferred over `ROS_TOPIC_STATS_QUIET` for stderr-parsing deployments: nothing is discarded, so a spec that failed to arm or an output file that can't be opened is still reported. |
```

This also composes with R6's planned "opt-in exporter as a **sidecar reading the log**" (docs/ROADMAP.md): a sidecar tailing the stats file plus the diagnostic file needs no probe-side network code at all.

Keep the QUIET row, but re-point it. `QUIET=1` is the right answer when the operator genuinely wants the banner gone and has nowhere to put a file (read-only rootfs, distroless image) — say that, rather than presenting it as the answer for stderr-parsing deployments in general.

test/integration/test_quiet.py
```diff
@@ -0,0 +1,99 @@
...
+    half-silence.
+    """
+    for v in ("0", "true", "yes"):
+        err = _run_capture(extra_env={"ROS_TOPIC_STATS_QUIET": v})
+        assert BANNER in err, (
+            f"QUIET={v!r} must NOT silence the banner (only '1' does)\n--- stderr ---\n{err}")
+
+
+def test_quiet_never_silences_error_diagnostics():
+    """QUIET=1 + malformed ROS_TOPIC_STATS_EXPECTED: the 'invalid spec' warning MUST still
```
TanayK07, line 86: ## 🔬 GAP ANALYSIS: the load-bearing invariant is claimed for four diagnostic paths and pinned for one

**Severity:** MEDIUM — a later change gating the flush-thread `cannot open output file` warning on `m_quiet` ships green. The invariant this file's own header calls load-bearing is 25% covered.

### The problem

Three places in this PR enumerate the diagnostics that must survive `QUIET=1`, all four each time:

- `interposers.cpp:139-140` — "unreadable/malformed/zero-rule spec in loadRateSpec, unwritable output file in flush"
- `README.md:112` — "unreadable/malformed/zero-rule spec, unwritable output file"
- `CHANGELOG.md:22-23` — same list

This test asserts one of them:

```python
assert "invalid ROS_TOPIC_STATS_EXPECTED" in err
```

`MALFORMED_SPEC = "bogus: 1\n"` drives `parseRateSpec` failure only. `readSpecFile` failure (`interposers.cpp:85`), the zero-rule branch (`interposers.cpp:102`), and the `fopen` failure in `flush()` (`interposers.cpp:345`) are unpinned.

### Why the three uncovered paths are more at risk than the covered one

1. **`cannot open output file` lives in a different function on a different thread.** `loadRateSpec()` sits next to `m_quiet`'s initialiser, so anyone touching it reads the QUIET rationale block. `flush()` is ~200 lines away and already carries its own once-guard (`m_warned_open_fail`, line 344) — precisely the shape ("we already suppress this to avoid spam") that invites someone to fold in `&& !m_quiet` during a cleanup. Nothing in the suite would fail.

2. **The zero-rule branch is the one whose real-world instance this PR cites as its justification** — CHANGELOG 0.2.0's directory-read-as-spec parsing to a zero-rule no-op. Asserting the rationale's own worked example is what turns the invariant from a statement of intent into a regression test.

3. `readSpecFile` failure is the cheapest to trigger and shares a known-good fixture with `test_rate_spec.py:221`, which already asserts `"not a regular file"` on the `pulse-check` side.

### Theory / proof

For a cross-cutting predicate — here "QUIET must not reach any error path" — coverage has to be per **call site**, not per predicate, because the predicate is re-implemented by omission at every site. It is the same reasoning behind `test/integration/test_symbols.py` in this repo enumerating the full export list instead of spot-checking one symbol, and behind the kernel's `CONFIG_FAULT_INJECTION` targeting individual allocation sites rather than a single representative one.

Stated as mutation testing (DeMillo/Lipton/Sayward 1978; Papadakis et al., *Mutation Testing Advances*, 2019): insert `&& !m_quiet` at `interposers.cpp:344` and the suite stays green. A surviving mutant on the line an invariant names is the definition of a missing assertion.

### Recommendation

Four cases, one parametrisation, roughly +6 s wall clock. All reuse machinery already present here or in `test_rate_spec.py`:

```python
import pytest

# (fixture kind, substring the diagnostic MUST contain) for every stderr write QUIET must not
# reach. Per-CALL-SITE, not per-message-kind: each fprintf(stderr, ...) in interposers.cpp is a
# separate opportunity to add `&& !m_quiet`.
@pytest.mark.parametrize("kind,expect", [
    ("malformed",  "invalid ROS_TOPIC_STATS_EXPECTED"),      # parseRateSpec fail   (interposers.cpp:93)
    ("unreadable", "cannot read ROS_TOPIC_STATS_EXPECTED"),  # readSpecFile fail    (interposers.cpp:85)
    ("zero_rule",  "declares no topics and no nodes"),       # empty spec           (interposers.cpp:102)
    ("bad_output", "cannot open output file"),               # fopen fail in flush  (interposers.cpp:345)
])
def test_quiet_never_silences_error_diagnostics(kind, expect):
    with tempfile.TemporaryDirectory() as d:
        env = {"ROS_TOPIC_STATS_QUIET": "1"}
        run_s, period = 3.0, "1.0"
        if kind == "malformed":
            env["ROS_TOPIC_STATS_EXPECTED"] = _write(d, "bad.yaml", MALFORMED_SPEC)
        elif kind == "unreadable":
            env["ROS_TOPIC_STATS_EXPECTED"] = d            # a directory: readSpecFile rejects it
        elif kind == "zero_rule":
            env["ROS_TOPIC_STATS_EXPECTED"] = _write(d, "empty.yaml", "# only a comment\n")
        else:
            # Non-existent directory -> fopen(ENOENT) on the FLUSH THREAD, so this one needs a
            # window to actually elapse. The warning is once-per-process (m_warned_open_fail),
            # so a single short window is enough.
            env["ROS_TOPIC_STATS_OUTPUT_FILE"] = os.path.join(d, "no", "such", "dir", "x.log")
            run_s, period = 4.0, "0.5"
        err = _run_capture(extra_env=env, run_s=run_s, period=period)
    assert expect in err, (
        f"QUIET=1 swallowed the {kind} diagnostic — that silently disables something the "
        f"operator configured, the exact 0.2.0 bug class\n--- stderr ---\n{err}")
    assert BANNER not in err, f"banner leaked through QUIET=1 ({kind})\n--- stderr ---\n{err}"
```

Two prerequisites, both small:

- `_run_capture` needs a `period` parameter to thread into `ph.make_env`.
- The `bad_output` case relies on `extra` overriding `ROS_TOPIC_STATS_OUTPUT_FILE`. That already works — `probe_harness.py:59` runs `env.update(extra)` last — but it is worth a one-line comment there, since it is now load-bearing rather than incidental.

src/probe/interposers.cpp
```diff
@@ -252,6 +265,10 @@ class ProbeRuntime {
           // graphs don't accrue a `TOPIC /x 0.000000` line every window. Set ROS_PULSE_EMIT_IDLE=1
           // to restore the legacy behaviour of printing them. See KNOWN_ISSUES.md #7.
           m_emit_idle(envFlag("ROS_PULSE_EMIT_IDLE")),
+          // Banner suppression for stderr-parsing deployments (ROADMAP R6). Same exact-"1"
+          // opt-in as every other flag: QUIET=true/yes/0 keep the banner, so a half-set flag
+          // fails loud (banner still there) instead of half-silencing.
+          m_quiet(envFlag("ROS_TOPIC_STATS_QUIET")),
```
TanayK07, line 281: ## 🔬 OPTIMIZATION: `envFlag()` belongs in the pure core — where it would get trimming and a microsecond unit test instead of a 9-second three-process one

**Severity:** MEDIUM — this PR is where the exact-`"1"` contract first gets documented and tested, and it tests it the most expensive way available while leaving the whole flag family inconsistent with the sibling env parsers on whitespace.

### The problem

`envFlag()` is an anonymous-namespace helper in the probe translation unit:

```cpp
// interposers.cpp:54-58
// True iff the env var is set to exactly "1" (the documented opt-in form).
auto envFlag(const char* key) -> bool {
    const char* v = std::getenv(key);
    return v != nullptr && v[0] == '1' && v[1] == '\0';
}
```

It has three callers now — `ROS_PULSE_EMIT_IDLE`, `ROS_TOPIC_STATS_JITTER`, and this line's `ROS_TOPIC_STATS_QUIET`. Because it lives inside `libros2_pulse.so`'s probe layer rather than `ros2_pulse::core`, none of the three has ever had a unit test, and this PR pays for that with `test_quiet_is_exact_opt_in` (test_quiet.py:72-84): three full `rclcpp` process launches at 3 s each — **~9 s of CI wall clock plus DDS discovery** — to assert three string comparisons.

Meanwhile the pure core already has exactly the right home. `src/core/env_config.cpp` holds `parsePeriodSeconds` and `parseMaxBytes`, and `test/unit/test_env_config.cpp` already pins their contract with six gtest cases including `ToleratesSurroundingWhitespace`.

### The second half: the trimming contract diverges

`env_config.cpp:31-32` states the project's position on operator input, and it is the opposite of `envFlag`'s:

```cpp
// Trim surrounding whitespace — operators export these via the shell, where a stray space is
// easy to introduce. Interior / trailing garbage is still rejected below.
```

Both core parsers trim. `envFlag` does not. So on the same env surface, in the same process:

- `ROS_TOPIC_STATISTICS_PUBLISH_PERIOD="2.0 "` → 2.0 s. Works.
- `ROS_TOPIC_STATS_MAX_BYTES=" 1048576"` → 1 MiB. Works.
- `ROS_TOPIC_STATS_QUIET="1 "` → banner still prints.
- `ROS_TOPIC_STATS_JITTER=" 1"` → **jitter silently stays off.**

The QUIET direction is benign and matches this comment's stated intent ("a half-set flag fails loud"). The `JITTER` direction is not benign: a trailing space in a docker-compose `environment:` block, a k8s ConfigMap value, or a `.env` file silently disables the measurement, and the log looks normal because gap tracking off simply omits `JITTER` lines. That is the same silently-disabled-measurement shape this PR argues against — arriving through whitespace instead of through QUIET.

Trailing whitespace in env values is a well-known operator hazard for exactly these formats: YAML plain scalars strip trailing spaces but block scalars and quoted forms do not, docker-compose `env_file` values are taken verbatim after the first `=` (Compose only strips quotes, not spaces), and `systemd` `EnvironmentFile=` preserves interior/trailing whitespace outside quotes. None of these are exotic.

Today all three flags work for anyone who types `=1` cleanly, which is nearly everyone, so nothing is broken. What breaks is the next reader: this PR's comment now asserts a deliberate exact-match *contract* for the family, and `test_quiet_is_exact_opt_in` pins `"true"`, `"yes"`, `"0"` — the values a reader would guess wrong on — but not `" 1"`, the value an operator actually produces by accident. The contract is documented at exactly the point where its most likely real violation is unpinned.

### Research

The two behaviours are both defensible in isolation; what is not defensible is having both on one surface. Prior art splits cleanly on *which* one you pick, never on mixing:

- glibc's tunables and `LD_*` boolean vars are strict presence/exact-value checks with no trimming — consistent across the family.
- Kubernetes and the OpenTelemetry SDK spec both normalise env values (OTel's spec mandates trimming and case-insensitive boolean parsing for all `OTEL_*` booleans, precisely to remove per-variable surprises).
- Go's `strconv.ParseBool` and Python's `distutils.util.strtobool` are the two poles; every codebase that survived picked one and applied it everywhere.

This project already picked: `env_config.cpp:31` is the written-down policy. `envFlag` is the one place it is not applied, and that is an accident of where the function happens to live.

### Recommendation

Move the helper into the core, next to the parsers whose policy it should share:

```cpp
// include/ros2_pulse/core/env_config.hpp
// True iff the value is exactly "1" after trimming surrounding whitespace. Exact-match by
// design: "true"/"yes"/"on" are NOT accepted, so a half-remembered flag fails loud (feature
// stays off, banner stays visible) rather than half-enabling. Trimming matches
// parsePeriodSeconds/parseMaxBytes — operators export these through shells, compose files and
// systemd EnvironmentFile=, all of which preserve a trailing space.
bool parseBoolFlag(const char* raw) noexcept;
```

```cpp
// src/core/env_config.cpp
bool parseBoolFlag(const char* raw) noexcept {
    if (raw == nullptr) return false;
    const char* first = raw;
    const char* last = raw;
    while (*last != '\0') ++last;
    while (first != last && isAsciiSpace(*first)) ++first;
    while (last != first && isAsciiSpace(*(last - 1))) --last;
    return (last - first) == 1 && *first == '1';
}
```

```cpp
// interposers.cpp — the probe keeps only the getenv shim
auto envFlag(const char* key) -> bool {
    return ros2_pulse::core::parseBoolFlag(std::getenv(key));
}
```

Then `test_quiet_is_exact_opt_in` becomes a gtest case in `test/unit/test_env_config.cpp` that runs in microseconds and covers all three flags at once:

```cpp
TEST(ParseBoolFlag, AcceptsOnlyExactOne) {
    EXPECT_TRUE(parseBoolFlag("1"));
    EXPECT_TRUE(parseBoolFlag(" 1 "));      // shell/compose/systemd trailing space
    EXPECT_FALSE(parseBoolFlag("true"));    // exact-match: half-set fails loud
    EXPECT_FALSE(parseBoolFlag("yes"));
    EXPECT_FALSE(parseBoolFlag("0"));
    EXPECT_FALSE(parseBoolFlag("11"));
    EXPECT_FALSE(parseBoolFlag(""));
    EXPECT_FALSE(parseBoolFlag(nullptr));
}
```

Net effect: ~9 s off every integration run, the string contract pinned in the fast lane where it belongs, `EMIT_IDLE` and `JITTER` get their first coverage for free, and the whitespace policy stops depending on which file a helper landed in. Keep one end-to-end QUIET case here (`QUIET=true` still shows the banner) so the wiring stays pinned — the exhaustive value matrix does not need a ROS process.

test/integration/test_quiet.py
```diff
@@ -0,0 +1,99 @@
...
+    multiple windows.
+    """
+    so, node = ph.probe_paths()
+    with tempfile.TemporaryDirectory() as d:
+        out = os.path.join(d, "quiet.log")
+        env = ph.make_env(so, out, period="1.0", extra=extra_env)
+        p = subprocess.Popen([node, "talker"], env=env,
+                             stdout=subprocess.DEVNULL, stderr=subprocess.PIPE, text=True)
+        try:
+            time.sleep(run_s)
```
TanayK07, line 48: ## 🔬 ISSUE: fixed `sleep(3.0)` × 6 process launches — the pass condition is "DDS came up fast enough", not "the banner printed"

**Severity:** MEDIUM — ~20 s of wall clock for four assertions, and `test_banner_present_by_default` fails with a misleading message whenever `rclcpp::init` + participant discovery overruns the sleep on a loaded runner.

### The problem

`_run_capture` waits a wall-clock constant, then kills the process and reads whatever accumulated:

```python
p = subprocess.Popen([node, "talker"], env=env, ...)
try:
    time.sleep(run_s)          # run_s = 3.0
finally:
    p.terminate()
```

Nothing here is conditioned on the banner. The docstring is right that the banner is emitted at attach time, so the *expected* latency is small — but the sleep encodes a bound on `rclcpp::init()` + `Node` construction, not on the probe. The four tests launch **six** processes total (`test_quiet_is_exact_opt_in` loops over three values), so the file costs ~18 s of sleep plus six ROS start/stop cycles regardless of outcome.

### Theory / what actually gates the banner

The banner fires from `ensureStarted()`, reached through `ros_trace_rcl_node_init` — the first tracepoint, inside `rcl_node_init`. Everything on the critical path to it is variable:

- `rclcpp::init` → `rcl_init` → `rmw_init`: participant creation. On Fast DDS this includes SHM segment creation and, unless `ROS_LOCALHOST_ONLY`/a limited-peers profile is set, UDP multicast socket setup on every interface.
- Under `colcon test`, several `ament_add_pytest_test` targets run concurrently by default, so these six launches contend with `integration_jitter` (240 s budget) and `integration_rate_spec` (180 s) for CPU and for the DDS discovery domain.
- Container/CI runners with a slow `/dev/shm` or many veth interfaces push participant creation into the hundreds of milliseconds to low seconds.

The margin is probably fine most of the time. The failure mode when it is not is the problem: `assert BANNER in err` reports *"default run lost the active banner"* — a QUIET regression — when the actual cause is that the node had not finished initialising. That misattribution is expensive to debug, and it is a false failure on the test whose whole job is to be the control.

Today, on an idle developer box, all six launches clear the banner in well under a second and everything passes. It breaks when the runner is loaded, when the CI image's DDS defaults change, or when someone raises `colcon test --parallel-workers`. Because `test_banner_present_by_default` is the *control* for the other three, an intermittent failure there also removes the evidence that `test_quiet_suppresses_banner`'s "no banner" result means anything — a suppressed banner and a banner that never got printed are indistinguishable from the assertion.

### Research

This is the standard sleep-versus-poll defect. The established guidance is to wait on the observable event with a generous deadline rather than on a fixed duration:

- pytest / flaky-test literature (Luo et al., *An Empirical Analysis of Flaky Tests*, FSE 2014) puts async-wait — "the test sleeps for a fixed time instead of waiting for the condition" — as the single largest flaky category, ~45% of the flaky tests studied.
- `launch_testing`, the ROS 2 first-party harness, provides `proc_output.assertWaitFor(expected_output=…, timeout=…)` for exactly this: assert on the stream, bounded by a timeout, returning as soon as it matches. This suite deliberately avoids `launch_testing` (`probe_harness.py:6-7`), so the equivalent has to be hand-rolled — but the semantics to copy are clear.
- Same reason `test/integration/test_stall.py` and the jitter tests here already reason in terms of *windows elapsed* rather than fixed sleeps where they can.

### Recommendation

Wait for the banner (or its absence) on a deadline. Reading stderr on a background thread keeps the pipe drained too, which removes the secondary risk that a chatty node fills the 64 KiB pipe buffer and blocks before `terminate()`:

```python
import threading

def _run_capture(extra_env=None, settle_s=1.0, deadline_s=15.0, period="1.0", await_text=BANNER):
    """Run one probed talker with stderr CAPTURED; return the decoded stderr text.

    Returns as soon as `await_text` appears (the positive tests) or after `settle_s` past node
    start with no match (the negative tests) — never a fixed sleep. The deadline bounds only the
    pathological case, so a loaded runner costs latency, not a false 'lost the active banner'.
    A drain thread also keeps the pipe from filling if the node ever gets chatty.
    """
    so, node = ph.probe_paths()
    with tempfile.TemporaryDirectory() as d:
        out = os.path.join(d, "quiet.log")
        env = ph.make_env(so, out, period=period, extra=extra_env)
        p = subprocess.Popen([node, "talker"], env=env, stdout=subprocess.DEVNULL,
                             stderr=subprocess.PIPE, encoding="utf-8", errors="replace")
        buf, seen = [], threading.Event()
        def drain():
            for line in p.stderr:
                buf.append(line)
                if await_text and await_text in line:
                    seen.set()   # match found; release the positive tests early
        t = threading.Thread(target=drain, daemon=True)
        t.start()
        try:
            # Positive case exits on the event; negative case still needs a settle window long
            # enough for the banner to have appeared if the feature were broken.
            if not seen.wait(timeout=deadline_s):
                time.sleep(settle_s)
        finally:
            p.terminate()
            try:
                p.wait(timeout=5)
            except subprocess.TimeoutExpired:
                p.kill()
                p.wait(timeout=5)
            t.join(timeout=5)
    return "".join(buf)
```

Negative tests then pass `await_text=None` so they take the settle path deliberately, and — importantly — should assert a **positive** anchor too, so "no banner" cannot be satisfied by "the node never started". The diagnostics test already does this correctly (it asserts `invalid ROS_TOPIC_STATS_EXPECTED` is present alongside `BANNER not in err`); `test_quiet_suppresses_banner` should get the same treatment, e.g. assert the stats file at `out` exists and contains at least one `# ts_ns=` block before concluding the silence was caused by QUIET rather than by a node that never attached.

Expected effect: the three `test_quiet_is_exact_opt_in` launches return in ~1 s each instead of 3 s, and the control test stops being a race.

test/integration/test_quiet.py
```diff
@@ -0,0 +1,99 @@
...
+    Short run: the banner and the spec diagnostics are both emitted at attach time (first
+    tracepoint / runtime construction), so we only need the process to start, not to flush
+    multiple windows.
+    """
+    so, node = ph.probe_paths()
+    with tempfile.TemporaryDirectory() as d:
+        out = os.path.join(d, "quiet.log")
+        env = ph.make_env(so, out, period="1.0", extra=extra_env)
+        p = subprocess.Popen([node, "talker"], env=env,
+                             stdout=subprocess.DEVNULL, stderr=subprocess.PIPE, text=True)
```
TanayK07, line 46: ## 🔬 ISSUE: `text=True` decodes a stream the test does not control, with `errors="strict"`

**Severity:** LOW — a single undecodable byte on the probed node's stderr turns every test in this file into a `UnicodeDecodeError` **error** instead of an assertion failure with the log attached, and the probe's own messages are the non-ASCII in question.

### The problem

```python
p = subprocess.Popen([node, "talker"], env=env,
                     stdout=subprocess.DEVNULL, stderr=subprocess.PIPE, text=True)
```

`text=True` with no `encoding=` wraps the pipe in a `TextIOWrapper` using `locale.getencoding()` and the default `errors="strict"`. Two properties of this call site make that the wrong default:

1. **Every line this file asserts on contains U+2014 EM DASH.** `interposers.cpp:154` (`active — interposing tracetools layer`) and all three spec diagnostics (`interposers.cpp:86, 94, 103`) embed it. In UTF-8 those are the bytes `E2 80 94`.
2. **The stream is not the probe's alone.** It is the whole node's fd 2 — `rclcpp` logging, the RMW implementation's own output, and anything Fast DDS/Cyclone writes. The test asserts on the probe's lines but decodes everyone's bytes.

Under an ASCII codec (`ANSI_X3.4-1968`), `E2` is undecodable and `communicate()` raises. Because the raise happens inside the `finally:` block's `communicate()` call, it propagates out of `_run_capture` as a test **error** — no captured stderr in the report, so the failure gives you nothing to debug with, and it hits all four tests at once.

Today it passes, and for a good reason worth stating: PEP 538 (Python 3.7+) coerces a `C`/`POSIX` `LC_CTYPE` to `C.UTF-8` when available, and PEP 540 UTF-8 Mode turns on automatically for those locales, so the usual ROS CI image decodes UTF-8 regardless of `LANG`. This is not a live failure.

It breaks where that coercion does not apply:

- `PYTHONCOERCECLOCALE=0` with `PYTHONUTF8=0` — set in some hardened/reproducible-build images specifically to keep locale behaviour pinned.
- A genuinely non-UTF-8, non-C locale. Latin-1 is harmless here (every byte maps, you just get mojibake and the ASCII `BANNER` substring still matches), but a strict single-byte codec that rejects `0x94`/`0x9D` — several ISO-2022 and EUC variants — raises.
- Any future non-UTF-8 byte from the node or the RMW layer, whatever the locale. This is the durable one: the test's correctness currently depends on nothing in the ROS stack ever emitting a byte the ambient codec rejects.

This file is the first place in the repo that decodes the *probe's* stderr — `test_fork.py:31` and `test_shutdown.py:54` use `stderr=DEVNULL`, and the `test_jitter.py`/`test_rate_spec.py` `text=True` calls target `pulse-check`, whose output the tests fully control. So the convention being set here is new, and it will be copied.

### Research

CPython's own guidance for subprocess pipes carrying foreign output is to name the codec and choose a non-strict error handler: `subprocess` documents `encoding`/`errors` as the explicit alternative to `text=True`'s locale default, and PEP 540's rationale is that "strict" on external byte streams converts data anomalies into crashes. `errors="replace"` is the standard choice for log scraping — U+FFFD is not silently wrong the way `errors="ignore"` is, since a substring assertion against a replaced region simply fails and the replacement characters are visible in the printed log. The same reasoning is why `git`'s Python tooling, `pytest`'s own capture layer, and `launch_testing`'s `proc_output` all decode with a replacement handler rather than strict.

### Recommendation

One-line fix; name the encoding, never let decoding decide the verdict:

```python
p = subprocess.Popen([node, "talker"], env=env,
                     stdout=subprocess.DEVNULL, stderr=subprocess.PIPE,
                     # Explicit codec + replacement handler: this pipe carries the whole node's
                     # stderr (rclcpp + RMW), not just ours, and every probe line we assert on
                     # contains U+2014. A strict locale codec would turn an unexpected byte into
                     # a UnicodeDecodeError with no captured log, instead of a failed assert.
                     encoding="utf-8", errors="replace")
```

Optionally assert the em dash survived once, so a future switch to a non-UTF-8 output encoding in `interposers.cpp` is caught rather than silently mojibaked past the ASCII `BANNER` check:

```python
BANNER = "[ros2_pulse] active"
BANNER_FULL = "[ros2_pulse] active — interposing tracetools layer"   # non-ASCII, pinned on purpose
```

CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ All notable changes to this project are documented here. Format follows
   `max_gap_ms` rule enables it implicitly so a declared rule is never silently unchecked.
 - `pulse-check` exits 2 when a spec requires a gap the logs cannot answer — measurement absence
   is not a health verdict. Violations it *did* measure are still printed.
+- **Quiet banner (ROADMAP R6):** `ROS_TOPIC_STATS_QUIET=1` suppresses the `[ros2_pulse] active`
```
TanayK07, line 20: ## 🔬 ISSUE: `docs/ROADMAP.md` still lists this feature as unshipped, and still calls R5 "next"

**Severity:** LOW — the roadmap is cited by commit messages and by this CHANGELOG entry as the stable index of feature state; after this PR it reports two features as planned that are already in `main`'s history.

### The problem

This entry announces "**Quiet banner (ROADMAP R6)**" as shipped. `docs/ROADMAP.md` disagrees, in two places:

```markdown
docs/ROADMAP.md:8-9
Current order of work: **R5** (next — it closes a soundness hole in the R1 alerting we
just shipped), then R6, then R2 if R5 hasn't already covered the need.

docs/ROADMAP.md:175
- `ROS_TOPIC_STATS_QUIET=1` to silence the stderr banner for stderr-parsing deployments.
```

R5 is not "next" — gap tracking is in `interposers.cpp` (`m_registry.setGapTracking(...)`, `has_pub_max_dt`/`has_recv_max_dt` in `window_format.cpp`), `test_jitter.py` is wired into `CMakeLists.txt`, and the Unreleased CHANGELOG section directly above this entry documents it as delivered. And line 175 is the feature this PR implements, still listed as a to-do bullet.

### Theory / why this is a factual error rather than a docs nit

The file's own header defines its contract:

> Section numbers are stable (referenced from commits and CHANGELOG), so they are **not** priority order.

That makes ROADMAP.md a referenced index, not prose — this very entry dereferences "ROADMAP R6", as does the PR title, the `CMakeLists.txt` comment, and the rationale block in `interposers.cpp`. An index whose entries misreport state breaks every reference into it: a contributor picking up work reads line 9, concludes R5 is unstarted, and duplicates shipped work; a reader following "ROADMAP R6" from this line lands on a bullet list that does not indicate which bullets are done.

The repo already established the convention that fixes this. R1 was closed out in place:

```markdown
docs/ROADMAP.md:11-16
## R1. Expected-rate spec + alerting — turns the logger into a monitor — **DONE**

> Shipped: `ROS_TOPIC_STATS_EXPECTED` spec (documented YAML subset, no YAML lib in the probe),
> flush-time `WARN TOPIC` / `WARN NODE` lines with first-window grace, and the no-ROS
> `pulse-check` CLI (exit 0/1/2) re-deriving verdicts from raw log rates. …
```

R6 is a multi-bullet section, so it cannot be marked `**DONE**` wholesale — which is precisely why per-bullet state is needed rather than optional.

### Recommendation

Mark the shipped bullet and correct the ordering line in the same commit that ships the feature, following the R1 precedent:

```markdown
## R6. Output & ecosystem (small, independent)

- `ROS_TOPIC_STATS_FORMAT=jsonl` — one object per window; golden-file tests for both formats.
- Callback names via `rclcpp_callback_register` for human-readable labels.
- Opt-in exporter (Prometheus/OTel) as a **sidecar reading the log** — keeps the probe itself
  network-zero.
- Validate on `rmw_zenoh` (no DDS at all) and CycloneDDS+iceoryx SHM; add support-matrix rows.
  Hooks sit above rmw, so both should work unmodified — worth proving.
- **DONE** — `ROS_TOPIC_STATS_QUIET=1` silences the stderr banner for stderr-parsing
  deployments. Scoped to the informational banner only; error diagnostics always print
  (see README, and `test/integration/test_quiet.py` for the pinned invariant).
```

and at line 8-9:

```markdown
Current order of work: **R5 shipped** (it closed the soundness hole in the R1 alerting), the
QUIET bullet of **R6** shipped; next is the rest of R6, then R2 if R5 hasn't covered the need.
```

Since three sources now cross-reference roadmap section numbers (commits, CHANGELOG, and the `interposers.cpp` rationale block), it is worth making "tick the roadmap bullet" an explicit line in the PR checklist rather than relying on it being remembered — the drift here happened across two consecutive feature PRs, which suggests it is a process gap rather than a one-off.

PR #28, opened 2026-08-08
test: probe + live LTTng session capture the same run (closes ROADMAP R3)

Tanay10x:
## What

The last open R3 item. The probe's whole compatibility claim with ros2_tracing rests on the `dlsym(RTLD_NEXT)` forwarding in every interposer, if it works, a live LTTng session and the probe are fed by the *same* tracepoint call. Nothing ever asserted it; a silently dropped forward (a typo'd dlsym name returning nullptr) would pass every other test in the suite.

`test_lttng_coexist.py`: start an LTTng session on `ros2:*`, run the 50 Hz talker with the probe preloaded, assert **both** consumers ate, probe log shows a plausible `/chatter` rate AND babeltrace2 (fallback babeltrace v1) reads >0 `ros2:*` events from the trace.

**Skip-vs-fail design** (where the assertion gets teeth): missing lttng CLI / dead sessiond / no viewer → skip (environment can't test). `ldd libtracetools.so` lacking lttng-ust → skip (stock Humble: the real tracepoint body is a no-op, 0 events proves nothing, the bake-off measured exactly 0 there). Environment fine but 0 events with the probe preloaded → **FAIL**: forwarding broke. An unprobed-talker baseline leg pins that the session machinery itself works.

**CI wiring:** rosdep keys `lttng-tools` + `babeltrace` as `test_depend` (verified in rosdistro base.yaml, no `babeltrace2` key exists; v1 reads LTTng's CTF 1.8). industrial_ci's rosdep step installs test_depends, so no ADDITIONAL_DEBS and no CI/local drift. Net: jazzy + kilted lanes RUN the test; humble installs the tools and self-skips on the ldd check, the documented, correct behavior there.

## Verification

- **ros:jazzy container: PASS**, probe log `TOPIC /chatter 50.002823` while babeltrace2 read **2,385 ros2:\* events** from the same run (251 `rcl_publish`, 250 `callback_start/end`, the interposed-and-forwarded tracepoints). Both consumers fed, neither starved. `rosdep install` in-container proved the key→apt chain (`lttng-tools 2.13.11`, `babeltrace 1.5.11`).
- **ros:humble container with lttng installed: SKIP** with the ldd reason (stock humble tracetools has no lttng-ust), exactly the CI-lane behavior.
- **Host (humble, no lttng): SKIP**, "lttng CLI not installed".
- Standalone fast lane 104/104, no regression. ROADMAP R3 ticked; CHANGELOG updated.


test/integration/test_lttng_coexist.py
```diff
@@ -0,0 +1,220 @@
...
+        # (a) The probe still measured a plausible rate under a live session. 30% tolerance
+        # matches the other integration tests (containers/CI schedulers are jittery).
+        ph.assert_rate_within(text, TOPIC, ph.KNOWN_RATE_HZ, field="topic",
+                              rel_tol=0.30, note="(with live LTTng session)")
+        events = sess.ros2_event_count()
+        # (b) Forwarding delivered events to lttng-ust. >0 is the claim; a 6 s 50 Hz talker
+        # actually produces hundreds of rcl_publish events alone, so 0 here — in an environment
+        # the skip-guards proved capable of tracing — means dlsym(RTLD_NEXT) forwarding is
+        # broken, which is precisely the regression this test exists to catch.
+        assert events > 0, (
```
TanayK07, line 195: ## 🔬 GAP ANALYSIS — `events > 0` over `ros2:*` cannot fail, even with every `dlsym(RTLD_NEXT)` forward removed

**Severity:** HIGH — the one assertion this PR exists to make does not discriminate the thing it claims to test. The test would stay green through the exact regression it is written to catch.

### The problem

Line 143 enables **every** tracepoint in the namespace:

```python
self._run("lttng", "enable-event", "-u", "ros2:*")
```

and line 195 asserts on the **total** count across that namespace:

```python
assert events > 0, ("live LTTng session captured 0 ros2:* events while the probe was preloaded — "
                    "interposer forwarding (dlsym RTLD_NEXT) is broken; ...")
```

But the probe only interposes a subset of the `ros2:*` tracepoints. `src/probe/interposers.cpp` exports exactly eight symbols:

```
ros_trace_rcl_node_init              ros_trace_rcl_publish
ros_trace_rcl_publisher_init         ros_trace_rclcpp_intra_publish
ros_trace_rcl_subscription_init      ros_trace_callback_start
ros_trace_rclcpp_subscription_init
ros_trace_rclcpp_subscription_callback_added
```

(confirmed by `grep '^ROS2_PULSE_EXPORT' src/probe/interposers.cpp`, and pinned by `test/integration/test_symbols.py` + `src/probe/exports.map`.)

### Theory / proof

`LD_PRELOAD` interposition is per-symbol. The dynamic linker only redirects calls whose symbol our `.so` defines; every other `ros_trace_*` call in `libtracetools.so` binds to the real body and reaches `lttng-ust` **without passing through any interposer**. This repo already documents the size of that gap — `docs/issues/issue-2-node-liveness.md:36-38`:

> "Verified against the Humble header `ros2/ros2_tracing @ humble : tracetools/include/tracetools/tracetools.h` — **all 30 declared tracepoints** are `*_init` (...) or runtime (`rcl_publish`, `callback_start`, **`callback_end`**, `rmw_take`, executor events, lifecycle transition)."

30 declared tracepoints, 8 interposed. The other 22 — including `callback_end`, `rmw_publish`, `rclcpp_publish`, `rcl_init`, `rcl_timer_init`, `rclcpp_timer_link_node`, `rclcpp_callback_register` — are emitted by `libtracetools` regardless of what our forwarding does.

**Sabotage test.** Replace every interposer body with a bare `return` (no `dlsym`, no forwarding at all). Re-run this test on jazzy:

- **(a)** `assert_rate_within` still passes — the probe's own counting is untouched.
- **(b)** the 6 s / 50 Hz talker still emits ~300 `callback_end`, ~300 `rmw_publish`, ~300 `rclcpp_publish` and the init events, none of which our probe touches. `events` is in the high hundreds. `assert events > 0` **passes.**

The PR description's own verification numbers prove the arithmetic: **2,385** `ros2:*` events, of which 251 `rcl_publish` + 250 `callback_start/end`. Even taking the most generous reading of that grouping, well over a thousand of those 2,385 events never traverse an interposer. Zeroing out the entire forwarding path removes a minority of the count and never approaches the `> 0` floor.

### Research

The repo's own bake-off gets this right — `bench/run_bakeoff.sh:94-95`:

```sh
# fair parity: enable only the two events our probe + eBPF hook
lttng enable-event -u 'ros2:rcl_publish','ros2:callback_start'
```

The file header claims this test "mirrors the leg-D sequence in bench/run_bakeoff.sh" (line 122), but it widened `'ros2:rcl_publish','ros2:callback_start'` to `ros2:*`. That widening is precisely what destroys the discriminating power, because the leg-D filter happened to be exactly the interposed set.

For comparison, `ros2_tracing`'s own `tracetools_test` asserts on **named** events with expected counts (`assertEventsSet`, `get_events_with_name`), never on a namespace total — for this reason.

### Recommendation

Enable only interposed tracepoints, and assert **per event name** with a floor derived from the run, not `> 0`:

```python
# Only tracepoints our probe interposes: if forwarding breaks, these go to zero and
# nothing else in the namespace can mask it.
INTERPOSED = ("ros2:rcl_publish", "ros2:callback_start")

def __enter__(self):
    self._run("lttng", "create", self.name, "--output", self.out_dir)
    self._run("lttng", "enable-event", "-s", self.name, "-u", ",".join(INTERPOSED))
    self._run("lttng", "start", self.name)
    return self

def event_counts(self):
    """{event_name: count} for the interposed tracepoints."""
    r = subprocess.run([_viewer(), self.out_dir], capture_output=True, text=True, check=False)
    return {name: sum(1 for ln in r.stdout.splitlines() if f"{name}:" in ln)
            for name in INTERPOSED}
```

and in the test — a 6 s run at 50 Hz publishes ~300 times; a floor of half the probe-measured count is robust to container jitter while still being orders of magnitude above the "forwarding removed" case (exactly 0):

```python
counts = sess.event_counts()
floor = int(0.5 * ph.KNOWN_RATE_HZ * 6.0)   # ~150
for name in INTERPOSED:
    assert counts[name] >= floor, (
        f"{name}: {counts[name]} events (< {floor}) with the probe preloaded — the "
        f"interposer for this symbol is not forwarding via dlsym(RTLD_NEXT).\n"
        f"all counts: {counts}\nprobe log:\n{text}")
```

Per-name counts also catch the narrower regression the current form misses entirely: **one** interposer losing its forward (a typo'd `dlsym` name in a single `realFn<...>("ros_trace_...")` call — the exact failure mode the file header names on line 12) while the other seven still work.

Worth verifying the fixed test actually fails: stub out `realFn` to return `nullptr` for `ros_trace_rcl_publish`, rebuild, and confirm the assertion fires. Until it has been seen red, the teeth are unproven.

test/integration/test_lttng_coexist.py
```diff
@@ -0,0 +1,220 @@
...
+        assert r.returncode == 0, (
+            f"{' '.join(argv)} failed (rc={r.returncode})\n"
+            f"stdout: {r.stdout}\nstderr: {r.stderr}")
+
+    def __enter__(self):
+        # Session creation was reachability-checked in _ensure_sessiond, so failures from here
+        # on are asserted, not skipped: an environment that can create sessions but not enable
+        # ros2:* events is broken in a way worth surfacing loudly.
+        self._run("lttng", "create", self.name, "--output", self.out_dir)
+        self._run("lttng", "enable-event", "-u", "ros2:*")
```
TanayK07, line 143: ## 🔬 ISSUE — `enable-event` / `start` omit `-s <session>`, so they act on global `.lttngrc` state; a home-less environment turns SKIP into a hard FAIL

**Severity:** MEDIUM — breaks the skip-vs-fail contract this file is built around, and is racy against any other `lttng create` by the same Unix user.

### The problem

```python
self._run("lttng", "create", self.name, "--output", self.out_dir)
self._run("lttng", "enable-event", "-u", "ros2:*")
self._run("lttng", "start")
```

`create` is named. `enable-event` and `start` are not. `__exit__` (lines 150-151) *is* named. The class already knows `self.name` on every line — it just doesn't pass it on two of them.

### Theory / proof

Per the LTTng 2.13 man page for `lttng-enable-event(1)`, without `--session`:

> "The current recording session (see `lttng-concepts(7)` to learn more about the current recording session)."

and on where that lives:

> "`$LTTNG_HOME/.lttngrc`: Unix user's LTTng runtime configuration. This is where LTTng stores the name of the Unix user's **current recording session between executions of `lttng(1)`**."

`$LTTNG_HOME` defaults to `$HOME`. So both unnamed commands resolve their target through a **file on disk, shared by every `lttng` invocation of that user**, not through the session handle this context manager holds.

Works today: one test process, a writable `$HOME`, serial ctest. Two ways it stops working:

**1. No writable `$HOME` → hard FAIL instead of SKIP.** `lttng create` talks to the sessiond and succeeds server-side, then tries to persist the name to `$LTTNG_HOME/.lttngrc`. If `HOME` is unset or read-only (rootless containers with no home entry, `docker run -u $(id -u)` against an image with no matching passwd row, some CI sandboxes), the next command finds no current session and errors out. That lands in `_run`'s `assert r.returncode == 0` (line 134) — a **failed** test, in an environment the file's own design says should skip. The header on lines 20-29 draws the skip/fail line at "environment can't test" vs "our forwarding broke"; this crosses it in the wrong direction, and the failure message will read as a probe regression.

**2. Concurrent `lttng create` by the same user retargets us.** `.lttngrc` is last-writer-wins across processes. Anything else running `lttng create` between our `create` and our `start` — a parallel ctest (`CTEST_PARALLEL_LEVEL`), a second colcon job on a shared runner, a developer in another terminal — silently moves `enable-event` and `start` onto *their* session. Ours then has zero enabled events and is never started, `ros2_event_count()` returns 0, and the test fails claiming `dlsym(RTLD_NEXT)` forwarding is broken. That is the single worst misdiagnosis this file can emit.

### Research

`lttng-tools`' own test suite (`tests/utils/utils.sh`) threads an explicit session name through every helper (`start_lttng_tracing_ok $SESSION_NAME`, `enable_ust_lttng_event_ok $SESSION_NAME ...`) rather than relying on the current-session pointer, for exactly this reason. `ros2_tracing`'s `tracetools_trace` drives the sessiond through the `lttng` Python bindings with an explicit session handle and never touches `.lttngrc`.

### Recommendation

Name the session on every command — two tokens, no behaviour change in the happy path:

```python
def __enter__(self):
    self._run("lttng", "create", self.name, "--output", self.out_dir)
    self._run("lttng", "enable-event", "-s", self.name, "-u", "ros2:*")
    self._run("lttng", "start", self.name)
    return self
```

And to make the skip contract hold rather than depend on it: point `LTTNG_HOME` at the test's own tempdir so `.lttngrc` is always writable and never shared. Passing it through `_run`'s env also isolates these tests from a developer's ambient LTTng state:

```python
def _run(self, *argv):
    env = dict(os.environ, LTTNG_HOME=self.lttng_home)   # tempdir created alongside out_dir
    r = subprocess.run(argv, capture_output=True, text=True, check=False, env=env)
    ...
```

(`_ensure_sessiond` and `_viewer` invocations need the same env, or the sessiond spawn will use a different `LTTNG_HOME` than the session commands and they will not find each other.)

test/integration/test_lttng_coexist.py
```diff
@@ -0,0 +1,220 @@
...
+        subprocess.run(["lttng", "destroy", self.name], capture_output=True, check=False)
+        return False
+
+    def ros2_event_count(self):
+        """Number of ros2:* events babeltrace can read back from the (stopped) session."""
+        r = subprocess.run([_viewer(), self.out_dir],
+                           capture_output=True, text=True, check=False)
+        # babeltrace exits non-zero on an empty/absent stream; that is a legitimate
+        # "0 events" outcome for the assertion, not a harness error.
+        return sum(1 for line in r.stdout.splitlines() if "ros2:" in line)
```
TanayK07, line 160: ## 🔬 ISSUE — the session is not scoped to the run, so the event count can be fed (or starved) by unrelated processes

**Severity:** MEDIUM — invalidates the baseline leg's control property, which is the stated source of the coexistence assertion's teeth.

### The problem

```python
def ros2_event_count(self):
    r = subprocess.run([_viewer(), self.out_dir], capture_output=True, text=True, check=False)
    return sum(1 for line in r.stdout.splitlines() if "ros2:" in line)
```

Nothing ties a counted event to the talker this test started. The count is "every `ros2:*` event any process wrote into this session".

### Theory / proof

An LTTng-UST recording session records **every instrumented application of the tracing user** that matches the enabled events — that is the whole point of the default per-UID buffering scheme (`lttng-enable-channel(1)`: `--buffers-uid` allocates "one set of ring buffers (one per CPU) **shared by all the instrumented processes**" of each Unix user). Process scoping is opt-in, via `lttng track --userspace --vpid <pid>` (default is `--all`, i.e. everything).

So the trace is a union over the machine, not a recording of our subprocess.

Today ctest is serial and the count happens to come from our talker. Three concrete ways that stops being true:

1. **`test_unprobed_talker_baseline` loses its control property.** Its docstring (lines 201-206) claims: *"if this passes and the probed run captured 0 events, the fault is provably in the probe's forwarding."* That inference requires the baseline's events to come from an un-probed process. Any other `lttng-ust`-linked ROS process alive on the box during those 4 seconds — a leftover `stats_probe_test_nodes` from an earlier test that outlived its `p.wait(timeout=5)`, a `ros2 daemon`, another package's integration test — contributes events. The baseline then passes on someone else's traffic and proves nothing.

2. **Parallel ctest inverts the diagnosis.** With `CTEST_PARALLEL_LEVEL > 1` (or two colcon jobs on one runner), `integration_accuracy`/`integration_jitter`/`integration_stall` all spawn talkers, and all of them are `lttng-ust`-linked on jazzy. Their events land in whichever session is recording. The probed test can then pass on a *neighbour's* events while our own forwarding is dead.

3. **Combined with the `ros2:*` filter** (see the comment on line 195), the failure mode compounds: a broad namespace filter plus an unscoped session means the count is dominated by events that are neither ours nor from an interposed symbol.

### Research

`ros2_tracing`'s `tracetools_test` runs each traced app in a dedicated session and reads back with `bt2` while filtering on the known process — it does not count a namespace across whatever else is running. LTTng's own docs present `lttng track --userspace --vpid` as the standard way to bound a session to a known process tree.

### Recommendation

Record `vpid` as a context and count only events from the process this test started. `lttng add-context` after `create`, before `start`:

```python
def __enter__(self):
    self._run("lttng", "create", self.name, "--output", self.out_dir)
    self._run("lttng", "enable-event", "-s", self.name, "-u", ",".join(INTERPOSED))
    self._run("lttng", "add-context", "-s", self.name, "-u", "-t", "vpid")
    self._run("lttng", "start", self.name)
    return self

def event_counts(self, pid):
    """{event_name: count} restricted to events emitted by `pid`."""
    r = subprocess.run([_viewer(), self.out_dir], capture_output=True, text=True, check=False)
    want = f"vpid = {pid}"
    return {name: sum(1 for ln in r.stdout.splitlines() if f"{name}:" in ln and want in ln)
            for name in INTERPOSED}
```

Both babeltrace2 and babeltrace v1 print enabled contexts inline in the default text sink (`..., { vpid = 1234 }, { ... }`), so the substring match works on either viewer.

This needs the talker's PID, which means `ph.run_single` has to surface it. Smallest change that keeps the existing harness contract: add an optional out-parameter or a sibling that returns `(text, pid)` —

```python
def run_single_with_pid(mode, out_path, run_s=6.0, period="1.0", extra_env=None, mode_args=()):
    """Same as run_single, but also returns the traced child's PID."""
```

For the baseline leg, `_run_unprobed_talker` already owns its `Popen`; just `return p.pid`.

Scoping to `vpid` also makes the assertion self-describing on failure: "0 events from pid N, 812 from other pids" is an immediately actionable message, where today's "0 events" is not.

test/integration/test_lttng_coexist.py
```diff
@@ -0,0 +1,220 @@
...
+        # stop before destroy so buffers are flushed to the CTF output; destroy even on test
+        # failure or sessions leak across test runs (lttng session names are daemon-global).
+        subprocess.run(["lttng", "stop", self.name], capture_output=True, check=False)
+        subprocess.run(["lttng", "destroy", self.name], capture_output=True, check=False)
+        return False
+
+    def ros2_event_count(self):
+        """Number of ros2:* events babeltrace can read back from the (stopped) session."""
+        r = subprocess.run([_viewer(), self.out_dir],
+                           capture_output=True, text=True, check=False)
```
TanayK07, line 157: ## 🔬 ISSUE — every babeltrace failure mode is silently recoded as "0 events", i.e. as "our forwarding is broken"

**Severity:** MEDIUM — a viewer/toolchain fault is reported to the developer as a probe regression, with the probe's log attached as if it were evidence.

### The problem

```python
r = subprocess.run([_viewer(), self.out_dir], capture_output=True, text=True, check=False)
# babeltrace exits non-zero on an empty/absent stream; that is a legitimate
# "0 events" outcome for the assertion, not a harness error.
return sum(1 for line in r.stdout.splitlines() if "ros2:" in line)
```

The comment states one reason a non-zero exit can occur, then the code discards `returncode` **and** `stderr` for *all* reasons. `stdout` empty for any cause becomes `0`, which line 195 converts into: *"interposer forwarding (dlsym RTLD_NEXT) is broken"*.

### Theory / proof

`babeltrace2(1)` exits non-zero on a wide class of conditions that have nothing to do with how many events were recorded:

- `src.ctf.fs` cannot find or parse `metadata` in the given directory (partial/corrupt trace, consumerd killed mid-flush, session destroyed before the flush completed)
- plugin path problems — `BABELTRACE_PLUGIN_PATH` unset in a stripped container, or the `babeltrace2-plugins-ctf` package absent while the `babeltrace2` binary is present (they are separate binaries on some distros)
- a CTF version the installed viewer cannot read
- permission errors on the output directory when the consumerd runs as a different uid than the test

Only the first bullet's degenerate case ("session recorded nothing at all") is the one the comment describes.

Today, in the ros:jazzy container the author verified, babeltrace2 is complete and the path is exercised only in its happy form.

The break: any CI image where the viewer is present-but-broken — which is reachable right now, because `_lttng_skip_reason()` (lines 85-90) probes for the viewer with `shutil.which()` only. `shutil.which("babeltrace2")` succeeds for a binary that cannot load its CTF plugin. The environment gate passes, the trace is unreadable, `events == 0`, and CI goes red with:

```
live LTTng session captured 0 ros2:* events while the probe was preloaded —
interposer forwarding (dlsym RTLD_NEXT) is broken; probe log was: ...
```

Someone then spends a day bisecting `interposers.cpp` for a bug that is an apt package.

This is a direct violation of the file's own stated contract (lines 20-29): *"environment can't test -> SKIP"*, *"environment fine but 0 events -> FAIL"*. The code cannot currently tell those two apart, so it reports the second in both cases.

### Research

`bench/run_bakeoff.sh:108` has the same shape (`babeltrace2 /work/trace 2>/dev/null | wc -l`) but it is a **benchmark that prints a number**, not an assertion — it reports `events captured (babeltrace2): $EVENTS` for a human to read, and lines 111-115 add an `ldd`-based note explaining a 0. Lifting the idiom into a hard `assert` is what makes swallowing the exit code unsafe.

### Recommendation

Distinguish "the trace genuinely has no events" from "the viewer could not read the trace", using the presence of CTF metadata as the discriminator:

```python
def ros2_event_count(self):
    """Events the viewer reads back. Raises if the trace exists but cannot be read —
    an unreadable trace is an environment fault, never evidence about our forwarding."""
    r = subprocess.run([_viewer(), self.out_dir], capture_output=True, text=True, check=False)
    if r.returncode != 0:
        # A session that recorded nothing may leave no metadata at all -> a true 0.
        # Metadata present + viewer failure -> the viewer is the problem, say so.
        has_metadata = any("metadata" in files
                           for _, _, files in os.walk(self.out_dir))
        if has_metadata:
            raise RuntimeError(
                f"{_viewer()} failed to read {self.out_dir} (rc={r.returncode}); the trace "
                f"exists but is unreadable, so this run says nothing about probe forwarding.\n"
                f"stderr: {r.stderr}")
    return sum(1 for line in r.stdout.splitlines() if "ros2:" in line)
```

`RuntimeError` (rather than `assert`) is deliberate: it surfaces as a pytest **error**, visually distinct from the assertion failure that means "the probe regressed", which matches the skip/fail/error taxonomy the header sets out.

A cheaper hardening on top: make `_viewer()` prove the binary works rather than merely exists, so this path degrades to a skip instead of an error —

```python
def _viewer():
    for exe in ("babeltrace2", "babeltrace"):
        path = shutil.which(exe)
        if path and subprocess.run([path, "--version"], capture_output=True,
                                   check=False).returncode == 0:
            return path
    return None
```

test/integration/test_lttng_coexist.py
```diff
@@ -0,0 +1,220 @@
...
+        # babeltrace exits non-zero on an empty/absent stream; that is a legitimate
+        # "0 events" outcome for the assertion, not a harness error.
+        return sum(1 for line in r.stdout.splitlines() if "ros2:" in line)
+
+
+def _run_unprobed_talker(run_s):
+    """Run the talker WITHOUT the probe (plain env) for run_s seconds — baseline leg."""
+    _, node = ph.probe_paths()
+    p = subprocess.Popen([node, "talker"], env=dict(os.environ))
```
TanayK07, line 166: ## 🔬 ISSUE — the "un-probed" baseline inherits `LD_PRELOAD` from the ambient environment, and never checks the node started

**Severity:** MEDIUM — two independent defects in a six-line helper whose entire job is to be a control.

### The problem

```python
def _run_unprobed_talker(run_s):
    """Run the talker WITHOUT the probe (plain env) for run_s seconds — baseline leg."""
    _, node = ph.probe_paths()
    p = subprocess.Popen([node, "talker"], env=dict(os.environ))
    try:
        time.sleep(run_s)
    finally:
        p.terminate()
        p.wait(timeout=5)
```

`dict(os.environ)` is described in the docstring as "plain env". It is not — it is *whatever the caller's environment happens to be*, copied verbatim.

### 1. `LD_PRELOAD` is inherited, so "un-probed" is not guaranteed

`README.md:93` documents the primary usage as a **shell export**:

```sh
export LD_PRELOAD=libros2_pulse.so      # resolved from the sourced workspace
```

A developer who followed the README in that terminal and then ran `colcon test` (or `pytest test/integration/test_lttng_coexist.py`) hands `LD_PRELOAD=libros2_pulse.so` straight through `os.environ` into this `Popen`. The baseline talker is then **probed**, and the leg silently stops being a control while still passing. The same applies to anything else that sets a process-wide preload — `/etc/ld.so.preload`, a sanitizer wrapper, `jemalloc`/`tcmalloc` shims common on robot images.

The harness right next door already models the correct contract: `probe_harness.make_env()` (line 52) *constructs* the env rather than inheriting it, and every probed test goes through it. The un-probed leg needs the mirror-image of that, not a raw copy.

Note this also silently inherits `ROS_TOPIC_STATS_OUTPUT_FILE` and `ROS_TOPIC_STATISTICS_PUBLISH_PERIOD`, which is harmless only *while* `LD_PRELOAD` happens to be absent.

### 2. No liveness check, so a dead node is reported as a broken LTTng environment

If `stats_probe_test_nodes` exits immediately — no RMW implementation resolvable, `ROS_DOMAIN_ID` misconfigured, a missing `libstd_msgs` after a partial install — `Popen` still succeeds (fork+exec worked), `time.sleep(4.0)` still elapses, `p.terminate()` on a dead pid is a no-op, and `p.wait()` returns instantly. Then:

```python
assert events > 0, (
    "un-probed talker produced 0 ros2:* events despite lttng-ust-linked tracetools — "
    "LTTng environment is broken in a way the skip-guards did not catch")
```

The message accuses LTTng. The actual fault is that the node never ran. `ph.run_single` guards against exactly this on line 129 (`assert os.path.exists(out_path), f"probe produced no output file for mode={mode}"`); the un-probed path has no equivalent because it produces no output file to check.

### Recommendation

Strip the preload explicitly and assert the child survived the run:

```python
def _run_unprobed_talker(run_s):
    """Run the talker WITHOUT the probe for run_s seconds — baseline leg.

    LD_PRELOAD is stripped explicitly, not merely assumed absent: README.md documents
    `export LD_PRELOAD=libros2_pulse.so` as the normal way to use this probe, so a
    developer's shell can otherwise turn this control leg into a second probed run.
    Returns the child's PID so the caller can scope the trace to this process.
    """
    _, node = ph.probe_paths()
    env = dict(os.environ)
    env.pop("LD_PRELOAD", None)
    p = subprocess.Popen([node, "talker"], env=env)
    try:
        time.sleep(run_s)
        assert p.poll() is None, (
            f"un-probed talker exited early (rc={p.returncode}) — the baseline leg never "
            f"produced traffic, so a 0-event result says nothing about LTTng or the probe")
    finally:
        p.terminate()
        p.wait(timeout=5)
    return p.pid
```

The `assert` inside `try` (before `finally`) is intentional: it fires while the process state is still meaningful, and `finally` still reaps.

Cheap follow-up that makes the control leg actually verifiable rather than assumed: have the baseline assert the *same* per-event counts the probed leg asserts (see the comment on line 195). Two legs asserting the same named-event floor — one probed, one not — is what turns "coexistence" from a `> 0` liveness check into a real comparison.

package.xml
```diff
@@ -21,6 +21,14 @@
...
   <test_depend>std_msgs</test_depend>
   <test_depend>ros2run</test_depend>
+  <!-- LTTng coexistence test (ROADMAP R3): drives a live LTTng session next to the probe and
+       reads the trace back. Both are verified rosdep keys in rosdistro's base.yaml (there is
+       NO `babeltrace2` key; the `babeltrace` key maps to the v1 CLI, which reads LTTng's
+       CTF 1.8 output fine — the test prefers babeltrace2 when the environment has it).
+       Installing them on every lane is deliberate: humble still skips via the ldd check on
+       libtracetools.so (no lttng-ust backend there), not for lack of tooling. -->
+  <test_depend>lttng-tools</test_depend>
+  <test_depend>babeltrace</test_depend>
```
TanayK07, line 31: ## 🔬 ISSUE — neither rosdep key resolves on RHEL, macOS, or openSUSE, so `rosdep install` now hard-fails there

**Severity:** MEDIUM — a test-only dependency breaks the *build* bootstrap for anyone on a non-Debian platform, including a ROS 2 Tier-3 platform.

### The problem

```xml
<test_depend>lttng-tools</test_depend>
<test_depend>babeltrace</test_depend>
```

The surrounding comment verifies the keys **exist** in rosdistro's `base.yaml`, and that there is no `babeltrace2` key. Both true. What was not checked is the **platform coverage inside those keys**.

From `rosdep/base.yaml` on `ros/rosdistro@master`:

```yaml
babeltrace:                        # lines 317-324
  arch: [babeltrace]
  debian: [babeltrace]
  fedora: [babeltrace]
  gentoo: [dev-util/babeltrace]
  nixos: [babeltrace]
  openembedded: [babeltrace@openembedded-core]
  ubuntu: [babeltrace]
                                   # no rhel, no osx/homebrew, no opensuse

lttng-tools:                       # lines 8246-8253
  debian: [lttng-tools]
  fedora: [lttng-tools]
  gentoo: [dev-util/lttng-tools]
  nixos: [lttng-tools]
  openembedded: [lttng-tools@openembedded-core]
  rhel: [lttng-tools]
  ubuntu: [lttng-tools]
                                   # no arch, no osx/homebrew, no opensuse
```

Note the gaps are not even the same set: `babeltrace` has `arch` but no `rhel`; `lttng-tools` has `rhel` but no `arch`. Neither has `osx` or `opensuse`.

### Theory / proof

`rosdep install` resolves **every** declared dependency for the detected OS before installing anything, and an unresolvable key is a fatal error, not a warning:

```
ERROR: the following packages/stacks could not have their rosdep keys resolved to system dependencies:
ros2_pulse: No definition of [babeltrace] for OS version [9]
```

`rosdep` has no per-platform conditional for the *resolution* step — `package.xml` `condition=` attributes key off environment variables (`$ROS_DISTRO`, `$ROS_VERSION`), not the host OS, so there is no manifest-level way to say "only on Debian derivatives".

Dependency **type** does not save this either: rosdep's default `--dependency-types` covers everything except `doc`, so `<test_depend>` is resolved by a plain `rosdep install --from-paths src --ignore-src -y`. (That is the same mechanism the PR relies on for industrial_ci to install these — it cuts both ways.)

CI is unaffected: all four lanes are Ubuntu-based, and `babeltrace` v1 is in universe on every relevant release (jammy 1.5.8, noble 1.5.11, questing 1.5.11, resolute 1.5.11), so the jazzy in-container verification in the PR description is genuine.

What breaks is the **source install** on:

- **RHEL 9 / Rocky / Alma** — `babeltrace` unresolvable. RHEL-9 is an official Tier-3 platform for both Humble and Jazzy, and this is a probe aimed at robot deployments, where RHEL images are not exotic.
- **macOS** — neither key has an `osx`/`homebrew` entry, so both fail. (Both tools do exist in Homebrew, the rosdep entries just don't list them.)
- **openSUSE** — neither key resolves.
- **Arch** — `lttng-tools` unresolvable.

In every case the user cannot get to a build at all, for two packages needed by a single test that would `pytest.skip` on most of those platforms anyway (the `ldd` guard, or the absent `lttng` CLI).

### Recommendation

Three options, in order of preference:

**1. Fix it upstream, keep the manifest honest.** Open a rosdistro PR adding the missing platform entries (`rhel`/`osx`/`opensuse` for `babeltrace`, `arch`/`osx`/`opensuse` for `lttng-tools`) — these are one-line additions each and routinely merged. Until it lands, ship option 2 or 3.

**2. Move the tooling to CI-only.** Drop both `<test_depend>` lines and install in the workflow instead, which also makes the humble lane stop installing tools it will never use:

```yaml
- uses: ros-industrial/industrial_ci@master
  env:
    ROS_DISTRO: ${{ matrix.ROS_DISTRO }}
    ADDITIONAL_DEBS: build-essential lttng-tools babeltrace
```

This gives up the "no CI/local drift" property the comment argues for. Given the test self-skips cleanly when the tools are absent, that property is worth less than a working `rosdep install` on four platform families.

**3. Keep them, document the escape hatch.** If the drift argument wins, say so where it costs someone an hour — in the README/CONTRIBUTING build instructions:

```sh
# On RHEL / macOS / openSUSE / Arch, the LTTng coexistence test's tooling has no
# rosdep entry; the test self-skips there, so skip the keys:
rosdep install --from-paths src --ignore-src -y --skip-keys "lttng-tools babeltrace"
```

Whichever path you take, the `package.xml` comment should record the platform gap alongside the "verified in base.yaml" claim — as written it reads as if coverage was checked, and the next person will trust it.

test/integration/test_lttng_coexist.py
```diff
@@ -0,0 +1,220 @@
...
+            f"{' '.join(argv)} failed (rc={r.returncode})\n"
+            f"stdout: {r.stdout}\nstderr: {r.stderr}")
+
+    def __enter__(self):
+        # Session creation was reachability-checked in _ensure_sessiond, so failures from here
+        # on are asserted, not skipped: an environment that can create sessions but not enable
+        # ros2:* events is broken in a way worth surfacing loudly.
+        self._run("lttng", "create", self.name, "--output", self.out_dir)
+        self._run("lttng", "enable-event", "-u", "ros2:*")
+        self._run("lttng", "start")
```
TanayK07, line 144: ## 🔬 ISSUE — the implicit UST channel sizes its ring buffers per CPU; on a many-core runner it exceeds Docker's default `/dev/shm` and the test hard-fails

**Severity:** MEDIUM — a CI failure whose cause (core count of the runner) is invisible from the error, on a test whose design says environment problems must skip.

### The problem

```python
self._run("lttng", "enable-event", "-u", "ros2:*")
self._run("lttng", "start")
```

No channel is configured, so `enable-event` creates the implicit `channel0` with defaults. `_run` asserts on a non-zero return code (line 134), so any allocation failure here is a **test failure**, not a skip.

### Theory / proof

From `lttng-enable-channel(1)` (v2.13), for the user space domain:

- default buffering scheme is `--buffers-uid` — "As of LTTng 2.13, this is the default buffering scheme for the user space tracing domain"
- `--buffers-uid` allocates "one set of ring buffers (**one per CPU**) shared by all the instrumented processes" of the user
- default sub-buffer size with `--buffers-uid`: **524288** bytes
- default sub-buffer count: **4**

So the shared-memory footprint is:

```
4 sub-buffers x 524288 B x nproc  =  2 MiB per CPU
```

LTTng-UST places these ring buffers in shared memory backed by `/dev/shm`. Docker's default `--shm-size` is **64 MB**, and `industrial_ci` does not override it.

| runner CPUs | UST shm needed | vs 64 MB default |
|---|---|---|
| 4 (GitHub `ubuntu-latest`) | 8 MiB | fine |
| 16 | 32 MiB | fine |
| 32 | 64 MiB | at the limit |
| 64 | 128 MiB | **fails** |

Today: GitHub-hosted `ubuntu-latest` is 4 vCPU, 8 MiB, comfortable. The author's local container verification would have been on a similar footing.

Breaks the moment the workload moves to bigger iron — a self-hosted runner, a larger GitHub runner label, or a maintainer running `colcon test` inside `docker run ros:jazzy` on a 32/64-core workstation. LTTng fails to allocate the buffers and `lttng start` (or the `enable-event` that creates the channel) returns non-zero. The `_run` assert then produces:

```
lttng start failed (rc=1)
stdout: ...
stderr: Error: Shared memory ... No space left on device
```

as a **red test**, when the file's own contract (lines 20-29) puts "environment can't test" firmly in the skip column. Worse, the failure is core-count dependent, so it reproduces on one machine and not another — the least debuggable class of CI flake.

The test needs a trivial amount of buffering: a 6 s talker at 50 Hz over two tracepoints is a few hundred KB of CTF total. It is paying for 2 MiB/CPU it will never use.

### Research

`lttng-tools`' own test suite sets small explicit channels rather than taking the defaults, precisely because CI machines vary in core count. `ros2_tracing`'s `tracetools_trace` exposes `--subbuf-size` and `--num-subbuf` on `ros2 trace` for the same reason, and the ROS 2 tracing docs recommend tuning them for constrained targets.

### Recommendation

Create the channel explicitly with a bounded size, then bind the events to it — the footprint becomes `4 x 64 KiB x nproc` = 256 KiB/CPU, a 32x reduction, still far more than this run produces:

```python
CHANNEL = "ros2_pulse"

def __enter__(self):
    self._run("lttng", "create", self.name, "--output", self.out_dir)
    # Explicit small channel: the implicit one defaults to 4 x 512 KiB per CPU
    # (lttng-enable-channel(1), --buffers-uid), which on a many-core runner exceeds
    # Docker's 64 MB default /dev/shm and turns an environment limit into a red test.
    # A 6 s 50 Hz talker over two tracepoints needs a tiny fraction of this.
    self._run("lttng", "enable-channel", "-s", self.name, "-u",
              "--subbuf-size", "64K", "--num-subbuf", "4", CHANNEL)
    self._run("lttng", "enable-event", "-s", self.name, "-u",
              ",".join(INTERPOSED), "-c", CHANNEL)
    self._run("lttng", "start", self.name)
    return self
```

Belt and braces, if you want the skip contract to hold rather than to depend on sizing: treat a failure of `enable-channel`/`start` specifically as a skip rather than an assert, since by that point `_ensure_sessiond` has already proven the daemon is reachable and a resource failure is unambiguously environmental:

```python
def _run(self, *argv, skip_on_failure=False):
    r = subprocess.run(argv, capture_output=True, text=True, check=False)
    if r.returncode != 0 and skip_on_failure:
        pytest.skip(f"{' '.join(argv)} failed (rc={r.returncode}): {r.stderr.strip()}")
    assert r.returncode == 0, (...)
```

Also worth noting for the discard-mode default: with `--num-subbuf 4` at 64K there is no realistic overrun here, but if the run length or the enabled-event set ever grows, add an assertion on the discarded-event count (`lttng stop` prints it) so a silently truncated trace can never be read as a low-but-nonzero pass.

test/integration/test_lttng_coexist.py
```diff
@@ -0,0 +1,220 @@
...
+        return "libtracetools.so not found on AMENT_PREFIX_PATH"
+    # The load-bearing check: is the REAL tracepoint body wired to lttng-ust? On stock humble
+    # tracetools is compiled with TRACETOOLS_DISABLED-style no-op bodies (not linked against
+    # lttng-ust), so an LTTng session captures nothing no matter what we forward — skipping
+    # there is correct and expected. Iron+ links the backend in the stock binaries.
+    out = subprocess.run(["ldd", lib], capture_output=True, text=True, check=False)
+    if "lttng-ust" not in out.stdout:
+        return (f"{lib} is not linked against lttng-ust (stock humble ships no-op "
+                "tracepoints; Iron+ required) — coexistence untestable here")
+    return None
```
TanayK07, line 77: ## 🔬 GAP ANALYSIS — nothing anywhere asserts this test ever actually runs, so a permanent silent skip is indistinguishable from a pass

**Severity:** MEDIUM — the failure mode of a skip-heavy test is not a red build, it is quiet irrelevance. This one has five independent paths to "skipped forever".

### The problem

Every guard in this file returns a skip reason, and each one swallows the tool's own failure:

```python
out = subprocess.run(["ldd", lib], capture_output=True, text=True, check=False)
if "lttng-ust" not in out.stdout:
    return (f"{lib} is not linked against lttng-ust ...")
return None
```

`check=False` and `returncode` unread. If `ldd` fails for *any* reason — the library path is stale, the file is a broken symlink after a partial install, `ldd` refuses a cross-arch object, SELinux denies the exec — `out.stdout` is empty, `"lttng-ust" not in ""` is `True`, and the test skips with a message asserting a fact about lttng-ust linkage that was never actually established.

The same shape appears in `_find_libtracetools` (returns `None` -> skip), `_lttng_skip_reason` (two skips), and `_ensure_sessiond` (skip). Five paths, all reporting `SKIPPED`, all in a suite where `colcon test` summarises skips as green.

### Theory / proof

`ament_add_pytest_test` reports a skipped pytest as a **passing** ctest test. `industrial_ci` with `AGGREGATE_TEST_RESULTS: 'true'` fails the job on test *failures*, not on skips. So the observable difference in CI between:

- jazzy running the test and proving forwarding works, and
- jazzy skipping the test because `ldd` moved, `AMENT_PREFIX_PATH` changed shape in a future distro, or the `babeltrace` deb got renamed

is **nothing**. Both are a green check.

That matters more here than for a normal test, because this file is the *only* thing in the repo that exercises the coexistence claim, and the PR closes ROADMAP R3 on the strength of it. The claim's verification quietly evaporating is exactly the outcome the roadmap item existed to prevent.

The `_find_libtracetools` path is a concrete near-term instance, not a hypothetical:

```python
cand = os.path.join(prefix, "lib", "libtracetools.so")
```

Unversioned `.so` in `<prefix>/lib` is a Debian-ROS packaging convention, not a guarantee. A merged-vs-isolated install layout change, or a distro that ships `libtracetools.so.4` with the unversioned symlink in a `-dev` package that is not installed in the runtime image, silently disables this test on every lane at once — with the message "libtracetools.so not found on AMENT_PREFIX_PATH", which nobody reads because the build is green.

### Research

This is a known failure mode in ROS 2 CI generally; `ros2_tracing`'s own `tracetools_test` gates on `TRACETOOLS_DISABLED`/`TRACETOOLS_LTTNG_ENABLED` compile-time state and its CI **asserts** that the tracing-enabled job ran the tracing tests rather than trusting the skip. `pytest`'s own guidance is the same: `--strict-markers`-style enforcement exists because conditional skips rot.

### Recommendation

Make "must run here" an explicit, environment-declared contract. The lanes that are supposed to run it declare it; the guards then fail loudly instead of skipping:

```python
def _skip_unless_testable():
    """Skip when the environment genuinely cannot test coexistence.

    ROS2_PULSE_REQUIRE_LTTNG=1 turns every skip into a failure. The jazzy/kilted CI lanes
    set it: on those images the tooling and the lttng-ust-linked tracetools are guaranteed,
    so a skip there means a guard silently stopped matching reality, not that the
    environment is incapable — and a silently-skipped test is indistinguishable from a
    passing one in an aggregated colcon report.
    """
    required = os.environ.get("ROS2_PULSE_REQUIRE_LTTNG") == "1"
    for reason in (_lttng_skip_reason(), _tracetools_skip_reason(), _ensure_sessiond()):
        if reason:
            if required:
                pytest.fail(f"ROS2_PULSE_REQUIRE_LTTNG=1 but coexistence is untestable: {reason}")
            pytest.skip(reason)
```

Wire it per lane in `.github/workflows/ci.yml`, which also documents the humble/jazzy split in the place CI readers look:

```yaml
- uses: ros-industrial/industrial_ci@master
  env:
    ROS_DISTRO: ${{ matrix.ROS_DISTRO }}
    # humble's stock tracetools has no lttng-ust backend, so the coexistence test
    # legitimately skips there. On Iron+ a skip means a guard broke, not that the
    # environment is incapable — make it fail instead of quietly passing.
    ROS2_PULSE_REQUIRE_LTTNG: ${{ matrix.ROS_DISTRO == 'humble' && '0' || '1' }}
```

(The `rolling` lane is `continue-on-error`, so setting it there is free signal.)

Separately, tighten the `ldd` guard so it stops reporting a conclusion it did not reach:

```python
out = subprocess.run(["ldd", lib], capture_output=True, text=True, check=False)
if out.returncode != 0:
    return f"ldd {lib} failed (rc={out.returncode}): {out.stderr.strip()}"
if "lttng-ust" not in out.stdout:
    return (f"{lib} is not linked against lttng-ust (stock humble ships no-op "
            "tracepoints; Iron+ required) — coexistence untestable here")
```

Two distinct reasons, two distinct messages — and under `ROS2_PULSE_REQUIRE_LTTNG=1` the first one becomes a failure that names the real cause.

test/integration/test_lttng_coexist.py
```diff
@@ -0,0 +1,220 @@
...
+def _ensure_sessiond():
+    """Make sure a session daemon is reachable; return a skip reason or None.
+
+    `lttng list` auto-spawns a per-user sessiond in most setups, but minimal containers can
+    lack that path — try an explicit daemonized spawn (--no-kernel: we only need the userspace
+    domain, and skipping the kernel domain avoids module-load noise as root) before giving up.
+    """
+    if subprocess.run(["lttng", "list"], capture_output=True, check=False).returncode == 0:
+        return None
+    subprocess.run(["lttng-sessiond", "--daemonize", "--no-kernel"],
```
TanayK07, line 102: ## 🔬 ISSUE — `lttng-sessiond` spawn raises `FileNotFoundError` (an ERROR, not a SKIP), and the daemon it starts is never reaped

**Severity:** LOW — narrow trigger, but both halves violate the skip contract the file's header sets out, and the fix is three lines.

### The problem

```python
if subprocess.run(["lttng", "list"], capture_output=True, check=False).returncode == 0:
    return None
subprocess.run(["lttng-sessiond", "--daemonize", "--no-kernel"],
               capture_output=True, check=False)
```

`check=False` suppresses a non-zero **exit status**. It does not suppress `FileNotFoundError`, which `subprocess` raises from the `exec` when the binary is not on `PATH` — before any return code exists.

### Theory / proof

`_lttng_skip_reason()` (line 86) gates on `shutil.which("lttng")` — the **client**. Nothing checks for `lttng-sessiond` — the **daemon**. These are the same upstream project but not always the same package:

- Debian/Ubuntu ship both in `lttng-tools` (so the rosdep key covers it), but
- Yocto/OpenEmbedded splits them (`lttng-tools` vs the sessiond binary in a separate package), and this repo targets embedded robot images — `bench/run_bakeoff.sh` and `test/orin/` both exist for that reason
- container images that install only the client to *read* remote traces are a normal minimal-image pattern

Reaching this line at all requires `lttng list` to have failed first, which is precisely the "something is odd about the LTTng install" state where a split package is most likely.

Result: `FileNotFoundError: [Errno 2] No such file or directory: 'lttng-sessiond'` propagates out of `_ensure_sessiond` -> `_skip_unless_testable` -> the test, and pytest reports **ERROR**. The header (lines 20-29) puts "sessiond won't start" explicitly in the SKIP column, and line 109 already has the correct message written for this case (`"lttng-sessiond could not be started/reached"`) — it is just unreachable when the binary is missing outright.

### Second half: the daemon outlives the suite

`--daemonize` forks a detached sessiond that nothing ever stops. `LttngSession.__exit__` destroys the *session*; the daemon persists after `pytest` exits, holding its `/dev/shm` ring buffers and its `$LTTNG_HOME` socket for the life of the container or the developer's login.

On CI that is harmless (the container dies). Locally it means running the suite leaves a background daemon a developer did not ask for, and a subsequent run inherits whatever state it accumulated — including, if a previous run crashed between `create` and `destroy`, a leaked session. Session names here are `pulse_coexist_{os.getpid()}`; PIDs recycle, so a leaked session can collide with a later run's name, `lttng create` returns non-zero, and `_run`'s assert turns it into a red test.

### Recommendation

Catch the missing binary and convert it to the skip reason that already exists, and make the spawn idempotent against a leaked session:

```python
def _ensure_sessiond():
    """Make sure a session daemon is reachable; return a skip reason or None."""
    if subprocess.run(["lttng", "list"], capture_output=True, check=False).returncode == 0:
        return None
    try:
        subprocess.run(["lttng-sessiond", "--daemonize", "--no-kernel"],
                       capture_output=True, check=False)
    except FileNotFoundError:
        # lttng client present without the daemon: some images split the packages.
        # This is an environment limitation -> skip, per the skip-vs-fail design above.
        return "lttng-sessiond not installed (lttng client present without the daemon)"
    for _ in range(10):
        if subprocess.run(["lttng", "list"], capture_output=True,
                          check=False).returncode == 0:
            return None
        time.sleep(0.5)
    return "lttng-sessiond could not be started/reached"
```

For the leaked-session half, make `__enter__` tolerant rather than asserting on a name collision:

```python
def __enter__(self):
    # A previous run killed between create and destroy leaves a same-named session
    # (PIDs recycle). Clear it rather than asserting on `lttng create`'s rc.
    subprocess.run(["lttng", "destroy", self.name], capture_output=True, check=False)
    self._run("lttng", "create", self.name, "--output", self.out_dir)
    ...
```

Pointing `LTTNG_HOME` at the test's tempdir (see the comment on line 143) subsumes most of this: each run gets its own daemon namespace, so leaked sessions and ambient developer state cannot reach these tests at all. If you take that route, spawn the sessiond with the same `LTTNG_HOME` and terminate it in a session-scoped fixture teardown — otherwise the daemon leak becomes a leak *per run* rather than one shared daemon.

docs/ROADMAP.md
```diff
@@ -101,10 +101,13 @@ humble / iron / jazzy / kilted / rolling, so the original "all exist on Humble"
   jazzy/kilted; full suite green on both.
 - ✅ `rclcpp_intra_publish` hooked → **publish-side intra** counts on Iron+ (additive `PUB`
   line; symbol exported-but-never-called on Humble).
-- ⬜ Remaining: integration assertion that the probe and a **live LTTng session** coexist
-  (we forward via `dlsym(RTLD_NEXT)`, so both should fire — assert it). Testable on any distro
-  whose binaries carry the lttng-ust backend: Iron+ out of the box, Humble only if ROS was
-  rebuilt for it.
+- ✅ Integration assertion that the probe and a **live LTTng session** coexist
```
TanayK07, line 104: ## 🔬 GAP ANALYSIS — R3 is marked DONE on the strength of an assertion that cannot currently fail

**Severity:** LOW (documentation), but it is the accuracy of the roadmap's central claim — contingent entirely on the finding at `test_lttng_coexist.py:195`.

### The problem

```markdown
- ✅ Integration assertion that the probe and a **live LTTng session** coexist
  (we forward via `dlsym(RTLD_NEXT)`, so both should fire — asserted by
  `test/integration/test_lttng_coexist.py`: probe log rates AND >0 `ros2:*` events in the same
  run, plus an un-probed baseline leg so a 0-event probed run is provably our forwarding).
```

and line 95: `## R3. Distro matrix: Jazzy / Kilted / rolling — **DONE**`.

The parenthetical is an accurate description of what the test does. The issue is that ">0 `ros2:*` events" does not establish the property the bullet claims.

### Theory / proof

The probe interposes 8 of the ~30 `ros2:*` tracepoints (`grep '^ROS2_PULSE_EXPORT' src/probe/interposers.cpp`; the 30 figure is this repo's own count in `docs/issues/issue-2-node-liveness.md:36-38`). The other ~22 — `callback_end`, `rmw_publish`, `rclcpp_publish`, `rcl_init`, `rcl_timer_init` among them — reach `lttng-ust` without passing through any interposer, because `LD_PRELOAD` interposition is per-symbol.

So a build with **every** `dlsym(RTLD_NEXT)` forward deleted still produces hundreds of `ros2:*` events for a 6 s 50 Hz talker, and `assert events > 0` passes. The forwarding claim R3 wanted verified is not yet verified by this test. Details, with the sabotage walkthrough and the arithmetic against this PR's own 2,385-event figure, are in the comment on `test/integration/test_lttng_coexist.py:195`.

Two smaller qualifiers on the same bullet:

- *"plus an un-probed baseline leg so a 0-event probed run is provably our forwarding"* — the baseline inherits `LD_PRELOAD` from the caller's environment (`_run_unprobed_talker`, line 166), and `README.md:93` documents `export LD_PRELOAD=libros2_pulse.so` as the normal way to run the probe. Run the suite from a README-configured shell and the control leg is itself probed. "Provably" is stronger than the current code supports.
- *"jazzy/kilted CI lanes run it"* — true when the guards match reality, but nothing asserts they did. A skipped pytest is a passing ctest, and `AGGREGATE_TEST_RESULTS` fails on failures, not skips. See the comment on line 77.

### Recommendation

Land the per-event-name assertion first (comment on line 195), then this bullet is accurate as written with one substitution: `>0 ros2:* events` becomes `per-tracepoint counts for the interposed symbols above a run-derived floor`.

If you would rather ship the test now and tighten later, downgrade the heading rather than the bullet, so the roadmap stays a reliable index of what is actually proven:

```markdown
## R3. Distro matrix: Jazzy / Kilted / rolling — **mostly DONE**

- 🔶 Integration assertion that the probe and a **live LTTng session** coexist —
  `test/integration/test_lttng_coexist.py` runs both consumers over the same run and
  asserts the probe's rates hold under a live session. The LTTng-side assertion is
  currently `>0 ros2:*` events, which does not yet isolate the ~8 interposed tracepoints
  from the ~22 that reach lttng-ust directly; tightening it to per-event counts is what
  closes this item.
```

The same qualifier applies to the CHANGELOG entry ("proving the `dlsym(RTLD_NEXT)` forwarding feeds both consumers") and to the header comment on `test_lttng_coexist.py:11-13` ("a silently dropped forward ... would pass every other test in this suite") — which is, as written, still true of this test too.

PR #29, opened 2026-08-08
test+docs: measured RMW matrix: CycloneDDS, iceoryx SHM, rmw_zenoh at 50 Hz on three distros (ROADMAP R6)

Tanay10x:
## What

ROADMAP R6: "Validate on `rmw_zenoh` (no DDS at all) and CycloneDDS+iceoryx SHM ... Hooks sit above rmw, so both should work unmodified, worth proving." Proved, and kept provable: `test/rmw/run_rmw_matrix.sh <distro>` is a one-command, re-runnable harness, stock `ros:<distro>` container, apt-installed RMWs, colcon-built probe, four legs (fastrtps control / cyclonedds / cyclonedds+iceoryx SHM via `iox-roudi`+XML / zenoh with its `rmw_zenohd` router), each leg = separate-process talker+listener + intra-process node with the probe preloaded, rates asserted 50 Hz ±30% by the same parser the CI accuracy suite trusts. Evidence (probe logs, stderr, daemon logs, exact deb versions) lands in `test/rmw/out/`.

## Results, all four configs green on humble AND jazzy AND kilted (2026-08-08)

- **CycloneDDS** (rmw 1.3.4/2.2.3/4.0.2): TOPIC 50.022 / RECV inter 50.020 / RECV intra 50.019 Hz
- **CycloneDDS + iceoryx SHM** (iceoryx-posh 2.0.5-2.0.6): 50.002 Hz both sides on a fixed-size UInt64 pair, `String` is not SHM-eligible, so a new `talker_fixed`/`listener_fixed` mode in `test_nodes.cpp` keeps the leg from silently measuring loopback UDP
- **rmw_zenoh** (0.1.9/0.2.9/0.6.6, Tier 1 on kilted): TOPIC 50.000 / RECV inter 50.001 / RECV intra 50.055 Hz
- fastrtps control: 50.007/50.002/50.055 Hz

## Findings stated, not papered over

- Stock `ros-<distro>-cyclonedds` debs already link `libiceoryx_posh` on all three distros, no rebuild caveat needed.
- Per-sample SHM-vs-loopback attribution is Cyclone-internal; the README footnote states exactly what the SHM run proves (RouDi attach + iceoryx endpoints on the measured topic in both processes + 50 Hz throughout) and what it can't.
- Harness self-finding: DDS multicast discovery crosses containers on the docker bridge, two concurrent runs read 100 Hz from two 50 Hz talkers. Fixed with per-distro `ROS_DOMAIN_ID`; the final humble+kilted runs were deliberately concurrent to prove the isolation.

README gains a "Middleware (RMW) matrix" section with the measured rows; ROADMAP R6 bullet ticked; CHANGELOG updated.


test/rmw/run_rmw_matrix.sh
```diff
@@ -0,0 +1,160 @@
...
+    ROUDI_PID=$!
+    sleep 2
+    if kill -0 "$ROUDI_PID" 2>/dev/null; then
+      CYCLONEDDS_URI="file:///tmp/cyclonedds_shm.xml" \
+        run_leg cyclonedds_shm rmw_cyclonedds_cpp --fixed
+      kill "$ROUDI_PID" 2>/dev/null; wait "$ROUDI_PID" 2>/dev/null
+      echo "---- roudi registrations (SHM runtime attach evidence) ----"
+      grep -iE "registered|application" "$OUT/cyclonedds_shm/roudi.log" | head -8
+      echo "---- cyclone shm trace (data-path evidence) ----"
+      grep -ilE "iceoryx|shm" "$OUT"/cyclonedds_shm/cdds-trace-*.log 2>/dev/null | head -4
```
TanayK07, line 104: ## 🔬 GAP ANALYSIS — the SHM leg's transport evidence is printed but never asserted, so `cyclonedds_shm` reports PASS even when iceoryx carried zero samples

**Severity:** HIGH — the one leg whose entire purpose is "don't silently measure loopback UDP" has no failing condition for silently measuring loopback UDP.

### The problem

Lines 101-104 collect exactly the right evidence and then throw it away:

```bash
echo "---- roudi registrations (SHM runtime attach evidence) ----"
grep -iE "registered|application" "$OUT/cyclonedds_shm/roudi.log" | head -8
echo "---- cyclone shm trace (data-path evidence) ----"
grep -ilE "iceoryx|shm" "$OUT"/cyclonedds_shm/cdds-trace-*.log 2>/dev/null | head -4
```

Neither `grep` feeds `FAILED`. Both are piped into `head`, so even their exit status is the pipeline's last command. The only thing that can fail this leg is `rmw_matrix_check.py`'s rate assertion — and a rate assertion cannot distinguish shared memory from UDP, because both deliver 50 Hz on a loopback-local 8-byte message.

### Theory / proof

Cyclone's iceoryx integration is **fail-open by design**. Per the Cyclone DDS 0.10 shared-memory docs (0.10.5 is the version on all three distros here — verified against `packages.ros.org`: `ros-humble-cyclonedds 0.10.5-2jammy`, `ros-jazzy-cyclonedds 0.10.5-1noble`, `ros-kilted-cyclonedds 0.10.5-2noble`), SHM requires *all* of: fixed-size type; `LIVELINESS_AUTOMATIC`; `DEADLINE = INFINITY`; `RELIABILITY_RELIABLE`; `DURABILITY_VOLATILE`; `HISTORY_KEEP_LAST` with depth ≤ the iceoryx publisher capacity (16). If **any** prerequisite is unmet — or RouDi is not reachable — "shared memory exchange is automatically disabled and data reverts to network transport". No error, no non-zero exit.

And upstream states the observational limit explicitly:

> "It is currently not possible to accurately check whether Shared Memory transfer actually takes place for a specific subscription."
> — [Cyclone DDS 0.10 Shared Memory](https://cyclonedds.io/docs/cyclonedds/0.10.2/shared_memory.html)

> "by just observing the received data it cannot be determined whether Shared Memory is actually used"
> — [ros2/rmw_cyclonedds `shared_memory_support.md`](https://github.com/ros2/rmw_cyclonedds/blob/master/shared_memory_support.md)

Observing received data is precisely and only what this leg does.

Today it likely passes honestly: RouDi comes up, the type is fixed-size, default rclcpp QoS (`KeepLast(10)`, RELIABLE, VOLATILE) satisfies every constraint, so SHM engages. The README footnote's quoted trace lines are real.

It breaks silently the moment any of these change, and none of them are hypothetical:
- someone gives `qos()` in `test_nodes.cpp` a `deadline` or switches it to `TransientLocal`+depth 20 → SHM disengages, leg still green;
- `<SharedMemory>` stops being the config form (see the Cyclone 0.11/PSMX comment on line 73) → SHM never engages, leg still green;
- RouDi comes up but its mempool config can't serve the segment → runtime falls back, leg still green.

In every case the README keeps claiming "✅ validated" for a row that measured UDP.

### Research

The two indicators upstream actually documents are (1) `iox-introspection-client --mempool` chunk usage rising to saturation, and (2) "if data rate or latency are beyond what is achievable using regular network communication this is an indication that ... Shared Memory ... is being used". (1) is ncurses-only (`ros-<distro>-iceoryx-introspection` exists as a deb — 2.0.5/2.0.6 for humble/jazzy/kilted — but its output is a TUI, so the README's "no scriptable introspection" caveat is fair). (2) *is* scriptable and is the standard trick in the Cyclone/iceoryx benchmarks.

### Recommendation

Two changes, both cheap:

**a) Turn the evidence you already gather into a gate** (after line 100):

```bash
      shm_ok=1
      grep -qiE "registered|application" "$OUT/cyclonedds_shm/roudi.log" || shm_ok=0
      for role in fixed_talker fixed_listener; do
        grep -q "DDS:Cyclone:rt/chatter_fixed" \
          "$OUT/cyclonedds_shm/cdds-trace-$role.log" 2>/dev/null || shm_ok=0
      done
      if [ "$shm_ok" -eq 0 ]; then
        echo "FAIL cyclonedds_shm: rates in band, but no RouDi attach + per-process iceoryx" \
             "endpoint for rt/chatter_fixed — SHM disengaged and this leg measured UDP"
        FAILED=1
      fi
```

(`fixed_talker` / `fixed_listener` are the `PULSE_ROLE` tags `rmw_matrix_check.py` assigns, so those are the exact trace filenames.)

**b) Add the negative control that makes the gate meaningful** — re-run the same fixed pair once with `CYCLONEDDS_URI` set but RouDi stopped, and assert the endpoint lines are *absent*. That is the difference between "we grep for a string that is always there" and "we grep for a string that is there iff SHM engaged", and it costs one extra 8 s leg.

With (a)+(b) the README footnote can stop hedging about attribution at the endpoint level — which is the level the row actually claims.

test/rmw/run_rmw_matrix.sh
```diff
@@ -0,0 +1,160 @@
...
+      || FAILED=1
+  }
+
+  run_leg fastrtps rmw_fastrtps_cpp
+  run_leg cyclonedds rmw_cyclonedds_cpp
+
+  # --- CycloneDDS + iceoryx shared memory ---
+  LIBDDSC="$(find "/opt/ros/${DISTRO}" -name 'libddsc.so.*' | head -1)"
+  ROUDI="$(command -v iox-roudi || find "/opt/ros/${DISTRO}" -name iox-roudi | head -1)"
+  if ldd "$LIBDDSC" | grep -q libiceoryx_posh && [ -n "$ROUDI" ]; then
```
TanayK07, line 73: ## 🔬 ISSUE — the iceoryx gate reads a Cyclone-0.10-only fact; on Cyclone 0.11 (already shipping in rolling) it reports "no iceoryx" for a distro that *does* support iceoryx

**Severity:** MEDIUM — false-negative SKIP that will make the matrix under-report SHM support on the next ROS distro, on a repo that already runs a rolling lane.

### The problem

```bash
LIBDDSC="$(find "/opt/ros/${DISTRO}" -name 'libddsc.so.*' | head -1)"
ROUDI="$(command -v iox-roudi || find "/opt/ros/${DISTRO}" -name iox-roudi | head -1)"
if ldd "$LIBDDSC" | grep -q libiceoryx_posh && [ -n "$ROUDI" ]; then
```

The gate encodes "iceoryx support == `libddsc` links `libiceoryx_posh`", and the config written at line 79-93 encodes "SHM is enabled via `<Domain><SharedMemory><Enable>`". Both are true for Cyclone DDS 0.10.x and false for 0.11.x.

### Theory / proof

Cyclone DDS 0.11 replaced the built-in iceoryx integration with **PSMX** (Pub/Sub Message Exchange): iceoryx moved out of `libddsc` into a loadable plugin (`psmx_iox`), and `<SharedMemory>` was replaced by a `<PubSubMessageExchange>` / PSMX interface element naming `library="psmx_iox"`. See [Shared memory configuration — Cyclone DDS latest](https://cyclonedds.io/docs/cyclonedds/latest/shared_memory/shared_mem_config.html); the zenoh-dds plugin tracked the same break in [eclipse-zenoh/zenoh-plugin-dds#212](https://github.com/eclipse-zenoh/zenoh-plugin-dds/issues/212) ("Cyclone DDS ... removed support for the previous Iceoryx shared memory implementation").

This is not a future concern — I checked `packages.ros.org` today:

```
ros-humble-cyclonedds   0.10.5-2jammy
ros-jazzy-cyclonedds    0.10.5-1noble
ros-kilted-cyclonedds   0.10.5-2noble
ros-rolling-cyclonedds  11.0.1-1noble      <-- PSMX era, no <SharedMemory>, no iceoryx in libddsc
ros-rolling-iceoryx-posh 2.0.5-5noble      <-- iceoryx is still there, just behind psmx_iox
```

Today (humble/jazzy/kilted, all 0.10.5) the gate is correct and the row is honest.

`test/rmw/run_rmw_matrix.sh rolling` — which the domain-ID comment on line 153 already anticipates ("rolling=42"), and which matches README's Rolling watch lane — takes the `else` branch and prints:

```
SKIP cyclonedds_shm: this distro's libddsc has no iceoryx linkage or no iox-roudi
```

That message is factually wrong for rolling: `ros-rolling-iceoryx-posh` is installed and iceoryx SHM works, just through `psmx_iox`. Whoever runs the matrix on the next distro will conclude SHM support regressed upstream. Worse, if someone "fixes" it by forcing the old branch, the 0.11 config parser rejects the unknown `<SharedMemory>` element rather than ignoring it, so the failure surfaces as an opaque participant-init/rate failure rather than "your config form is obsolete".

### Recommendation

Detect the era instead of one artifact of it, and make the skip message state which era it found:

```bash
  CDDS_VER="$(dpkg-query -W -f '${Version}' "ros-${DISTRO}-cyclonedds" 2>/dev/null)"
  PSMX_IOX="$(find "/opt/ros/${DISTRO}" -name 'libpsmx_iox.so*' | head -1)"
  if [ -n "$LIBDDSC" ] && ldd "$LIBDDSC" 2>/dev/null | grep -q libiceoryx_posh; then
    SHM_STYLE=sharedmemory          # Cyclone 0.10.x
  elif [ -n "$PSMX_IOX" ]; then
    SHM_STYLE=psmx                  # Cyclone 0.11+ (rolling: 11.0.1)
  else
    SHM_STYLE=none
  fi
```

and emit the matching config for `psmx`:

```xml
  <Domain Id="any">
    <SharedMemory><Enable>true</Enable></SharedMemory>   <!-- 0.10 form -->
    <!-- 0.11+ form: -->
    <PSMX><PSMXInstance name="iox" library="psmx_iox"/></PSMX>
  </Domain>
```

Two smaller things on the same three lines: `LIBDDSC` is unquoted-empty-safe but `ldd ""` still runs and spews to stderr when `find` misses, so guard with `[ -n "$LIBDDSC" ]` first; and `find ... | head -1` picks an arbitrary match if a distro ever ships two `libddsc.so.*` — prefer the one under `/opt/ros/$DISTRO/lib`.

test/rmw/run_rmw_matrix.sh
```diff
@@ -0,0 +1,160 @@
...
+    run_leg zenoh rmw_zenoh_cpp
+  else
+    echo "FAIL zenoh: router (rmw_zenohd) never opened tcp/7447 — see zenoh/router.log"
+    FAILED=1
+  fi
+  kill "$ROUTER_PID" 2>/dev/null; wait "$ROUTER_PID" 2>/dev/null
+
+  # Evidence must belong to the host user, not root — otherwise the next run cannot clean up
+  # its own out/ dir. /pulse_src is the host checkout, so its owner IS the host user.
+  chown -R "$(stat -c %u:%g /pulse_src)" "$OUT" 2>/dev/null || true
```
TanayK07, line 134: ## 🔬 ISSUE — the `chown` that keeps evidence host-owned only runs on the success path; every early `exit 2` leaves root-owned files in the user's checkout

**Severity:** MEDIUM — a failed run poisons the host tree in a way the user cannot clean up without `sudo`, and `.gitignore` hides it.

### The problem

```bash
chown -R "$(stat -c %u:%g /pulse_src)" "$OUT" 2>/dev/null || true
```

This is the last statement before `exit "$FAILED"`, so it is reached only if the container side runs to completion. But the container side has three unconditional early exits *after* it has already written into the bind-mounted `/pulse_out`:

- line 37 `apt-get update -qq || exit 2`
- line 40 `... apt-get install ... > "$OUT/apt.log" ... || { tail -5 "$OUT/apt.log"; exit 2; }` — writes `apt.log` as root, then exits
- line 52-53 `colcon build ... > "$OUT/build.log" ... || { tail -30 "$OUT/build.log"; exit 2; }` — same for `build.log`

`versions.txt` (line 42-48) is also written before any of those can fail.

### Theory / proof

The container runs as uid 0 by default (`docker run` with no `--user`), and `-v "$OUT_DIR:/pulse_out"` is a bind mount: files created inside are created on the host filesystem with the container's uid. The host user then owns the *directory* (created at line 147 by `mkdir -p`) but not the files inside it.

Happy path: fine, the `chown` fixes ownership.

Failure path — and apt failure is the *most likely* failure of this script, because it depends on `packages.ros.org` being reachable and on `ros-${DISTRO}-rmw-zenoh-cpp` existing for the requested distro: the user is left with `test/rmw/out/<distro>/apt.log` owned by `root:root` inside their own repo. The next run's `mkdir -p` succeeds (the dir is theirs), but any attempt to clear stale evidence — `rm -rf test/rmw/out`, a `git clean -xdf`, a CI cache wipe — fails with `EPERM`. Because line 30-32 of `.gitignore` hides `test/rmw/out/`, `git status` will not show why the tree is unclean.

### Recommendation

Make it a trap, so it runs on *every* exit from the container side. Place it immediately after `OUT=/pulse_out` (line 21):

```bash
  OUT=/pulse_out
  # Evidence must belong to the host user on EVERY exit path (apt failure, build failure,
  # SIGINT), not just the happy one — otherwise a failed run leaves root-owned files in the
  # host checkout that the user cannot clean up, and .gitignore hides them.
  trap 'chown -R "$(stat -c %u:%g /pulse_src)" "$OUT" 2>/dev/null || true' EXIT
```

and delete lines 132-134. `trap ... EXIT` in bash fires for normal exit, `exit N`, and (with `set -u` unaffected) for the shell being terminated by SIGINT/SIGTERM once you also `trap - EXIT`-style chain them — adding `INT TERM` to the trap spec covers the Ctrl-C case too.

test/rmw/run_rmw_matrix.sh
```diff
@@ -0,0 +1,160 @@
...
+for d in "${DISTROS[@]}"; do
+  OUT_DIR="$REPO_DIR/test/rmw/out/$d"
+  mkdir -p "$OUT_DIR"
+  echo "############ ros:$d ############"
+  # Distinct ROS_DOMAIN_ID per distro: DDS multicast discovery DOES cross containers on the
+  # default docker bridge, so concurrent matrix runs (or any domain-0 ROS traffic on the host)
+  # would feed a leg's listener from a foreign talker — observed as RECV 100 Hz from two 50 Hz
+  # talkers when the humble and kilted runs overlapped. cksum keeps the mapping deterministic
+  # (humble=66 jazzy=22 kilted=79 rolling=42) so evidence stays reproducible.
+  DOM=$(( ( $(printf '%s' "$d" | cksum | cut -d' ' -f1) % 90 ) + 10 ))
```
TanayK07, line 154: ## 🔬 ISSUE — the per-distro `ROS_DOMAIN_ID` fixes cross-*distro* interference only; two concurrent runs of the *same* distro collide 100% of the time, reproducing exactly the bug this line was added for

**Severity:** MEDIUM — the documented mitigation is deterministic in the wrong direction, and a stronger, upstream-supported mitigation exists.

### The problem

```bash
DOM=$(( ( $(printf '%s' "$d" | cksum | cut -d' ' -f1) % 90 ) + 10 ))
```

`cksum` is a pure function of the distro name, so the domain is a pure function of the distro name. That is deliberate ("cksum keeps the mapping deterministic ... so evidence stays reproducible") and it is exactly what makes the isolation partial:

- `run_rmw_matrix.sh jazzy` in one terminal and `run_rmw_matrix.sh jazzy` in another → **both get domain 22**, both containers on the default bridge, multicast discovery crosses → the listener sees two 50 Hz talkers on `/chatter` and reads 100 Hz. That is the identical artifact the comment above says was fixed.
- `run_rmw_matrix.sh humble jazzy humble` (retrying one distro) → same thing.
- Any host-side or third-party ROS 2 process on the bridge that happens to use domain 10-99 (a colleague's container, a `ros2 topic pub` left running, another checkout of this repo) → same thing, and it is invisible in the evidence because the foreign talker never writes to `test/rmw/out/`.

The failure mode is silent-in-the-wrong-direction: it inflates rates, and `measured_rate()` takes the **max** over interior windows, so an intermittent foreign talker is more likely to be captured than averaged away.

### Theory / proof

Domain ID does not isolate anything; it only changes the multicast group/port pair (`7400 + 250*domain + ...` in RTPS). Two participants with the same domain on a shared L2 segment — which is what the docker bridge is — discover each other by design. The docker bridge forwards multicast between containers on the same network, which is exactly what the comment observed.

ROS 2 has a first-class knob for this that does not depend on picking a lucky number: [Improved Dynamic Discovery](https://docs.ros.org/en/kilted/Tutorials/Advanced/Improved-Dynamic-Discovery.html) introduced `ROS_AUTOMATIC_DISCOVERY_RANGE` with values `SUBNET` (default for DDS RMWs — discover anything reachable by multicast), `LOCALHOST` (same machine only) and `OFF`. Iron deprecated the older `ROS_LOCALHOST_ONLY` in its favour. Since every talker/listener pair in this harness lives inside a single container, `LOCALHOST` is exactly the right semantics — and it bounds discovery by the container's network namespace rather than by hoping no one else picked domain 22.

`rmw_zenoh` needs no equivalent: its key expressions are `<domain_id>/<fqn>/<type>/<hash>` and upstream states sessions on different `ROS_DOMAIN_ID` cannot communicate even through shared routers, and UDP multicast scouting is off by default ([rmw_zenoh design.md](https://github.com/ros2/rmw_zenoh/blob/rolling/docs/design.md)). So this change only has to be right for the DDS legs.

### Recommendation

Belt and braces — keep the deterministic domain for evidence readability, but stop relying on it for isolation:

```bash
  # LOCALHOST discovery bounds every leg to its own container netns, so a concurrent run of the
  # SAME distro (identical DOM by construction) or any foreign domain-N traffic on the docker
  # bridge cannot feed a listener. ROS_AUTOMATIC_DISCOVERY_RANGE is Iron+; humble still needs
  # the deprecated predecessor.
  case "$d" in
    humble) DISC=(-e ROS_LOCALHOST_ONLY=1) ;;
    *)      DISC=(-e ROS_AUTOMATIC_DISCOVERY_RANGE=LOCALHOST) ;;
  esac
  docker run --rm --shm-size=1g -e ROS_DOMAIN_ID="$DOM" "${DISC[@]}" \
    -v "$REPO_DIR:/pulse_src:ro" -v "$OUT_DIR:/pulse_out" \
    "ros:$d" bash /pulse_src/test/rmw/run_rmw_matrix.sh --inside "$d" || RC=1
```

If you want the isolation to be structural rather than configured, the alternative is `docker run --network none` for the leg phase — loopback still exists, every leg is localhost-only by construction — but that has to come after the apt/colcon phase, so it means splitting the container run in two. The env-var route above gets ~all of the benefit for two lines.

Worth also recording the effective `ROS_DOMAIN_ID` and discovery range into `versions.txt`; right now the evidence bundle does not state under which isolation the numbers were taken.

test/rmw/rmw_matrix_check.py
```diff
@@ -0,0 +1,125 @@
...
+    out_l = os.path.join(capture_dir, f"{tag_prefix}listener.pulse.log")
+    pt, ft = _spawn(mode_t, out_t, capture_dir, f"{tag_prefix}talker")
+    pl, fl = _spawn(mode_l, out_l, capture_dir, f"{tag_prefix}listener")
+    try:
+        time.sleep(run_s)
+    finally:
+        for p in (pt, pl):
+            p.terminate()
+        for p in (pt, pl):
+            p.wait(timeout=10)
```
TanayK07, line 63: ## 🔬 ISSUE — `p.wait(timeout=10)` with no `kill()` fallback leaks a live 50 Hz talker into the next leg, which is the same double-count failure the harness already documents

**Severity:** MEDIUM — a stuck node does not just fail its own leg, it corrupts every leg that runs after it in the same container.

### The problem

```python
    finally:
        for p in (pt, pl):
            p.terminate()
        for p in (pt, pl):
            p.wait(timeout=10)
        ft.close()
        fl.close()
```

`subprocess.Popen.wait(timeout=...)` does not kill on timeout — it raises `TimeoutExpired` and leaves the child running. Raised inside a `finally`, it propagates out of `run_pair_modes`, out of `main()`, and out of the process (uncaught), so:

1. `ft`/`fl` never close (minor — CPython's refcounting will flush them);
2. the surviving talker keeps publishing at 50 Hz on `/chatter` in the container's `ROS_DOMAIN_ID`;
3. `run_leg` sees a non-zero exit and sets `FAILED=1` for *that* leg — correct;
4. the container continues to the next `run_leg`, whose fresh listener now receives from **two** talkers.

The same pattern is at line 112 for the intra leg (`pi.wait(timeout=10)`).

### Theory / proof

The termination path is not hypothetical. `p.terminate()` is SIGTERM; the node is `rclcpp::spin(node)` in `test_nodes.cpp`, which relies on rclcpp's default signal handler to break the spin and return. Under the SHM leg that shutdown additionally tears down an iceoryx runtime attached to RouDi — and this script `kill`s RouDi at line 100 right after the leg, so a slow-exiting process can find its RouDi gone mid-teardown. Ten seconds is generous, but "generous timeout" and "no recovery when the timeout is hit" are different properties.

Note the outcome is exactly the artifact the host-side comment (line 149-153) describes: *"observed as RECV 100 Hz from two 50 Hz talkers"*. The container-crossing instance of that bug was found and fixed; this in-container instance has the same signature and would be much harder to attribute, since both talkers are the harness's own.

### Research

The stdlib's own guidance for this is `Popen.communicate`'s documented recipe — on `TimeoutExpired`, `kill()` then `wait()` again ([subprocess docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate)). `test/integration/probe_harness.py:127-128` has the identical gap (`p.terminate(); p.wait(timeout=5)`), so a fix belongs there too — it is the shared harness this file deliberately reuses.

### Recommendation

Add a helper and use it in both call sites here (and, ideally, in `probe_harness.run_single`/`run_pair`):

```python
def _stop(p, tag, grace=10):
    """SIGTERM, then SIGKILL — never leave a 50 Hz publisher alive for the next leg."""
    p.terminate()
    try:
        p.wait(timeout=grace)
    except subprocess.TimeoutExpired:
        print(f"WARN {tag} ignored SIGTERM after {grace}s; SIGKILL", flush=True)
        p.kill()
        p.wait(timeout=5)
```

```python
    finally:
        _stop(pt, f"{tag_prefix}talker")
        _stop(pl, f"{tag_prefix}listener")
        ft.close()
        fl.close()
```

A SIGKILLed probe loses its exit flush, so the leg will fail its rate assertion — which is the right outcome, and it fails *only* that leg.

test/rmw/rmw_matrix_check.py
```diff
@@ -0,0 +1,125 @@
...
+import sys
+import time
+
+sys.path.insert(
+    0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "integration"))
+import probe_harness as ph  # noqa: E402
+
+# Same band as test/integration/test_accuracy.py: wide enough for wall-timer jitter + container
+# scheduling noise, tight enough to catch a double-count (~2x) or a dropped transport (0).
+TOL = 0.30
```
TanayK07, line 37: ## 🔬 GAP ANALYSIS — `±30%` on the *max* interior window is a liveness check, not a comparison metric; the README publishes it as a per-RMW steady-state rate

**Severity:** MEDIUM — the numbers in the matrix are a best-case sample, and the band is wide enough that a middleware losing a quarter of its samples still reads "✅ validated".

### The problem

```python
TOL = 0.30
```

combined with what the reused harness returns:

```python
# probe_harness.py:101-112
def measured_rate(windows, topic, field):
    interior = windows[1:-1] if len(windows) >= 3 else windows
    ...
    return max(vals) if vals else None
```

So each row asserts: *the single best 1 s window, excluding first and last, was within [35, 65] Hz.*

### Theory / proof

The comment on `TOL` is right about what the band was designed for — "wide enough for wall-timer jitter + container scheduling noise, tight enough to catch a double-count (~2x) or a dropped transport (0)". That is the correct band for `test_accuracy.py`, whose job is "is the probe's arithmetic right on the CI-tested RMW".

This file's job is different. `--leg zenoh` vs `--leg cyclonedds` is a *comparison across middlewares*, and the two statistics do not survive the change of purpose:

- **`max` over interior windows** is the sample maximum of ~6 draws. If a transport drops 25% of samples in half the windows, the max is still ~50 Hz. The estimator is chosen to be robust against ramp-up/truncation under-count, which makes it structurally blind to intermittent under-delivery — the most plausible way an unfamiliar RMW misbehaves.
- **±30%** means a leg reading 37 Hz — a 26% loss — prints `PASS ... 37.000 Hz (expect 50.0 +/-30%)`. The README would then quote a number the run itself did not consider notable.

Concretely: nothing in this harness would fail if `rmw_zenoh`'s router dropped every fourth sample under load, or if the iceoryx mempool ran short and Cyclone fell back per-sample. Both are exactly the RMW-specific behaviours the matrix exists to detect.

Today all legs land at 50.00-50.06 Hz, i.e. the real spread is ~0.1%, three orders of magnitude inside the band — so the band is not currently doing any work, and the max-vs-mean distinction does not change any published number. That is precisely why it is worth tightening now: the assertion is only ever consulted on the day something is wrong.

### Recommendation

Keep the reused parser (that reuse is the right call and is the whole "same parser CI trusts" argument), but make the leg report and assert a distribution:

```python
def _interior(text, topic, field):
    wins = ph.parse_windows(text)
    interior = wins[1:-1] if len(wins) >= 3 else wins
    return [v for v in (ph._value(w, topic, field) for w in interior) if v is not None]
```

then per check, alongside the existing `assert_rate_within`:

```python
    vals = _interior(text, topic, field)
    mean = sum(vals) / len(vals)
    sd = (sum((v - mean) ** 2 for v in vals) / max(1, len(vals) - 1)) ** 0.5
    print(f"[{args.leg}] {what}: n={len(vals)} mean={mean:.3f} sd={sd:.3f} "
          f"min={min(vals):.3f} max={max(vals):.3f}", flush=True)
    # a middleware that under-delivers intermittently shows up here, not in max()
    if min(vals) < ph.KNOWN_RATE_HZ * 0.90:
        failures.append(f"{what}: worst window {min(vals):.3f} Hz < 45 Hz")
```

Two benefits: the evidence bundle then contains the spread the README rows imply, and the README can quote `mean ± sd (n=6)` instead of a bare number whose estimator is `max`.

test/rmw/run_rmw_matrix.sh
```diff
@@ -0,0 +1,160 @@
...
+    <Tracing>
+      <Category>shm</Category>
+      <OutputFile>/pulse_out/cyclonedds_shm/cdds-trace-${PULSE_ROLE}.log</OutputFile>
+    </Tracing>
+  </Domain>
+</CycloneDDS>
+EOF
+    "$ROUDI" > "$OUT/cyclonedds_shm/roudi.log" 2>&1 &
+    ROUDI_PID=$!
+    sleep 2
```
TanayK07, line 96: ## 🔬 ISSUE — `sleep 2` + `kill -0` proves RouDi is *alive*, not that it is *serving*; a slow RouDi turns the SHM leg into a silent UDP leg

**Severity:** MEDIUM — a timing-dependent false PASS on the leg whose whole point is transport attribution.

### The problem

```bash
    "$ROUDI" > "$OUT/cyclonedds_shm/roudi.log" 2>&1 &
    ROUDI_PID=$!
    sleep 2
    if kill -0 "$ROUDI_PID" 2>/dev/null; then
```

`kill -0` answers "does this pid exist". RouDi's startup is not instantaneous: it creates and maps the management segment plus every mempool in its layout (which is why the host side passes `--shm-size=1g`), and only then announces readiness. Two seconds is a guess, and on a loaded CI box or a slower filesystem-backed `/dev/shm` it is a guess that can be wrong while the process is still very much alive.

### Theory / proof

If the applications start before RouDi is serving, iceoryx's runtime registration does not hard-fail the app — Cyclone logs the wait and, per the 0.10 shared-memory docs, "if prerequisites aren't met, shared memory disengages automatically" and traffic reverts to the network stack. `rmw_cyclonedds`'s own doc names the exact line you would see: `[Warning]: RouDi not found - waiting`.

So the failure is: RouDi alive but not ready → both nodes fall back to loopback UDP → both still publish/receive at 50 Hz → `rmw_matrix_check.py` passes → the leg is labelled `cyclonedds_shm`. This compounds with the missing evidence assertion (see the comment on line 104): there is no second gate to catch it.

RouDi prints an explicit readiness line (`RouDi is ready for clients` in iceoryx 2.0.x, which is what these distros ship — `ros-humble/kilted-iceoryx-posh 2.0.5`, `ros-jazzy-iceoryx-posh 2.0.6`), and it goes to the log file this script already captures.

### Recommendation

Poll the log for readiness instead of sleeping, with the same bounded-retry shape the zenoh leg already uses at line 120:

```bash
    "$ROUDI" > "$OUT/cyclonedds_shm/roudi.log" 2>&1 &
    ROUDI_PID=$!
    ROUDI_UP=0
    for _ in $(seq 1 40); do            # 20 s ceiling, same budget as the zenoh probe
      grep -q "RouDi is ready for clients" "$OUT/cyclonedds_shm/roudi.log" && { ROUDI_UP=1; break; }
      kill -0 "$ROUDI_PID" 2>/dev/null || break     # died: stop waiting, report below
      sleep 0.5
    done
    if [ "$ROUDI_UP" = 1 ]; then
      ...
```

This also improves the existing failure message: today `iox-roudi died at start` is printed for both "died" and "was too slow", which are different diagnoses (`/dev/shm` too small vs. host under load).

Same class of nit two blocks down: after `kill "$ROUDI_PID"` (line 100) and `kill "$ROUTER_PID"` (line 130), the unconditional `wait` has no ceiling — a daemon that ignores SIGTERM hangs the whole matrix run with no output. `timeout 10 tail --pid="$PID" -f /dev/null || kill -9 "$PID"` (or a simple bounded `kill -0` poll then `kill -9`) keeps the run bounded.

test/rmw/run_rmw_matrix.sh
```diff
@@ -0,0 +1,160 @@
...
+    echo "no-iceoryx" > "$OUT/cyclonedds_shm.skipped"
+  fi
+
+  # --- rmw_zenoh: bring the router up first ---
+  mkdir -p "$OUT/zenoh"
+  ros2 run rmw_zenoh_cpp rmw_zenohd > "$OUT/zenoh/router.log" 2>&1 &
+  ROUTER_PID=$!
+  ROUTER_UP=0
+  for _ in $(seq 1 40); do
+    if (exec 3<>/dev/tcp/127.0.0.1/7447) 2>/dev/null; then ROUTER_UP=1; break; fi
```
TanayK07, line 121: ## 🔬 ISSUE — the zenoh readiness probe tests "TCP accept", not "router usable", and leaks a file descriptor each time it succeeds

**Severity:** LOW — mostly cosmetic today, but the failure it cannot see is a discovery-timeout that would present as an unexplained 0 Hz.

### The problem

```bash
  for _ in $(seq 1 40); do
    if (exec 3<>/dev/tcp/127.0.0.1/7447) 2>/dev/null; then ROUTER_UP=1; break; fi
    sleep 0.5
  done
```

Two things:

1. **What it proves.** A successful `connect()` on 7447 means the listener socket exists and the backlog accepted a connection. `rmw_zenohd` binds its endpoint early in startup; a session that connects during that window can still fail to complete the handshake or receive no gossip. The probe's own success also *establishes and abandons* a TCP session against the router, which the router will log and time out — harmless, but it means the router log the evidence bundle keeps contains a phantom peer that never sent a Zenoh handshake.
2. **The fd.** `exec 3<>...` runs inside `( )`, so the descriptor lives in the subshell and dies with it — that is actually fine. But because it is a subshell, the connection is also never shut down cleanly (no `exec 3>&-`), so the router sees an abrupt RST rather than a close. Using `exec 3<>...; exec 3>&-` inside the subshell makes the probe's own footprint clean.

### Theory / proof

`rmw_zenoh` is router-mandatory by default: UDP multicast scouting is disabled ("to avoid issues with misconfigured networks, operating systems, or containers") and peers rely on the router's gossip scouting to find each other ([rmw_zenoh design.md](https://github.com/ros2/rmw_zenoh/blob/rolling/docs/design.md)). So "router up" is a precondition for the leg producing any data at all — which makes it worth checking the router's own readiness statement rather than its socket. `rmw_zenohd` logs its listening endpoints on startup into the file this script already captures (`$OUT/zenoh/router.log`).

Today the leg passes and reads 50.000/50.001/50.055 Hz, so the router was clearly ready. The scenario that breaks: a heavily loaded host where the router accepts on 7447 but has not finished initialising its gossip state when the talker/listener start. The pair then spends part or all of the 8 s window undiscovered, the listener reports well under 50 Hz, and the failure message points at *rates*, not at the router — sending the next reader of `test/rmw/out/` to debug the probe instead of the harness.

### Recommendation

Check both, cheaply — socket first (fast negative), then the router's own log line:

```bash
  for _ in $(seq 1 40); do
    if (exec 3<>/dev/tcp/127.0.0.1/7447; exec 3>&-) 2>/dev/null &&
       grep -qiE "listening on|zenohd.*(started|ready)" "$OUT/zenoh/router.log"; then
      ROUTER_UP=1; break
    fi
    kill -0 "$ROUTER_PID" 2>/dev/null || break   # router exited: fail fast with its log
    sleep 0.5
  done
```

(Adjust the pattern to whatever `rmw_zenohd` prints on the oldest distro you support — humble ships `rmw-zenoh-cpp 0.1.9`, so it is worth grepping the captured `router.log` from the humble run rather than assuming the rolling format.) The `kill -0` guard also converts "router crashed immediately" from a 20 s wait into an instant, correctly-attributed failure.

README.md
```diff
@@ -243,11 +243,39 @@ always linked to lttng-ust; then the tracepoints are no-ops. See [`bench/RESULTS
...
+
+All four configurations passed on all of `ros:humble`, `ros:jazzy` and `ros:kilted` on
+2026-08-08 (quoted rates are the jazzy run; humble and kilted matched within ±0.1 Hz — full
+per-distro logs under `test/rmw/out/`):
+
+| RMW | Status | Evidence |
+|---|---|---|
+| `rmw_fastrtps_cpp` (the default on all three distros) | ✅ CI-tested | every integration lane runs it; kept as the matrix control leg — TOPIC 50.007 / RECV inter 50.002 / RECV intra 50.055 Hz |
+| `rmw_cyclonedds_cpp` | ✅ validated | rmw 1.3.4 (humble) / 2.2.3 (jazzy) / 4.0.2 (kilted), cyclonedds 0.10.5: TOPIC 50.022 / RECV inter 50.020 / RECV intra 50.019 Hz |
+| CycloneDDS + iceoryx SHM | ✅ validated¹ | iceoryx-posh 2.0.5–2.0.6: `iox-roudi` up, `<SharedMemory><Enable>true</>`, fixed-size UInt64 pair (String is not SHM-eligible): TOPIC 50.002 / RECV inter 50.002 Hz |
```
TanayK07, line 269: ## 🔬 GAP ANALYSIS — the matrix rows are a dated, unpinned, unrepeated measurement: nothing in CI re-runs them, and `apt-get install` floats

**Severity:** MEDIUM — the README states a support claim that decays silently, on a repo whose other support claims are all CI-enforced.

### The problem

The rows read `✅ validated` with exact versions and a date. What produces them is `test/rmw/run_rmw_matrix.sh`, which is not referenced anywhere in `.github/workflows/ci.yml` — the workflow has `build-and-test` (industrial_ci over humble/jazzy/kilted), `standalone` (3 sanitizer lanes) and the rolling non-blocking lane, and none of them invoke `test/rmw/`. The container side also installs unpinned:

```bash
DEBIAN_FRONTEND=noninteractive apt-get install -y -qq \
  "ros-${DISTRO}-rmw-cyclonedds-cpp" "ros-${DISTRO}-rmw-zenoh-cpp"
```

so a re-run months from now measures a different stack than the one the table cites, and only `versions.txt` (inside gitignored `test/rmw/out/`) records which.

### Theory / proof — how fast this actually moves

Not a theoretical decay. Checking `packages.ros.org` today:

```
ros-humble-cyclonedds    0.10.5-2jammy      ros-humble-rmw-zenoh-cpp  0.1.9-1jammy
ros-jazzy-cyclonedds     0.10.5-1noble      ros-jazzy-rmw-zenoh-cpp   0.2.9-1noble
ros-kilted-cyclonedds    0.10.5-2noble      ros-kilted-rmw-zenoh-cpp  0.6.6-1noble
ros-rolling-cyclonedds   11.0.1-1noble   <-- major break: <SharedMemory> replaced by PSMX
```

`rmw_zenoh` moved 0.1.9 → 0.6.6 across three distros, and Cyclone has already made a breaking config change in rolling that this harness's iceoryx gate does not understand (see the comment on `run_rmw_matrix.sh` line 73). Both are the kind of change that turns a green row red — or, worse given the missing SHM assertion, keeps it green while it stops meaning what it says.

The rest of this repo does not accept that risk: every distro row in the *support* matrix above is backed by a blocking `build + test` job, and the Rolling row is explicitly labelled "🟡 non-blocking CI lane ... observational — watches upstream churn". The RMW matrix is the only claim in the README with no lane at all behind it.

### Recommendation

Add a fourth, non-blocking, scheduled lane — same status the Rolling row already carries, so it costs nothing in required checks:

```yaml
  rmw-matrix:
    name: rmw matrix (${{ matrix.ROS_DISTRO }}, non-blocking)
    if: github.event_name != 'pull_request'
    continue-on-error: true
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        ROS_DISTRO: [humble, jazzy, kilted]
    steps:
      - uses: actions/checkout@v4
      - run: test/rmw/run_rmw_matrix.sh ${{ matrix.ROS_DISTRO }}
      - uses: actions/upload-artifact@v4
        if: always()
        with: { name: rmw-out-${{ matrix.ROS_DISTRO }}, path: test/rmw/out/ }
```

with `on: schedule: - cron: '0 6 * * 1'` and `workflow_dispatch` added to the workflow triggers. Weekly is enough for apt churn, the artifact keeps the evidence the README points at, and `continue-on-error` means upstream churn shows as a red-but-non-blocking lane rather than a broken PR.

Two smaller hardening items that make a re-run comparable to this one:
- record `docker image inspect --format '{{index .RepoDigests 0}}' ros:$d` into `versions.txt` — the deb versions alone do not pin the base image;
- print the `versions.txt` delta against the versions quoted here, so a drifted run says *which* package moved instead of just failing a rate assertion.

test/rmw/run_rmw_matrix.sh
```diff
@@ -0,0 +1,160 @@
...
+  {
+    echo "image=ros:${DISTRO}  run_date=$(date -u +%F)"
+    dpkg-query -W -f '${Package} ${Version}\n' \
+      "ros-${DISTRO}-rmw-fastrtps-cpp" "ros-${DISTRO}-rmw-cyclonedds-cpp" \
+      "ros-${DISTRO}-cyclonedds" "ros-${DISTRO}-rmw-zenoh-cpp" \
+      "ros-${DISTRO}-iceoryx-posh" 2>/dev/null
+  } | tee "$OUT/versions.txt"
+
+  echo "==== colcon build (probe + test nodes) ===="
+  mkdir -p /ws/src && cp -r /pulse_src /ws/src/ros2_pulse
```
TanayK07, line 50: ## 🔬 OPTIMIZATION — `cp -r /pulse_src` copies `.git`, plus any host `build/`/`install/`, into the package source tree for every distro

**Severity:** LOW — wasted time per leg, and a real (if uncommon) way to get a confusing colcon failure that has nothing to do with the RMW under test.

### The problem

```bash
mkdir -p /ws/src && cp -r /pulse_src /ws/src/ros2_pulse
```

`/pulse_src` is the whole host checkout mounted read-only, so this copies everything the mount exposes: `.git/` (full history and objects), and — because they exist on the host but are gitignored, not absent — `build/`, `install/`, `log/` from any host-side `colcon build`, plus `test/rmw/out/` from previous matrix runs.

### Theory / proof

Two costs:

1. **Time and disk, multiplied by distros.** The copy runs once per container, i.e. once per distro argument. For a repo with meaningful history that is hundreds of MB of `.git` objects copied before every build, inside a container whose only need is the source tree.
2. **Correctness risk.** A host `build/` directory carries `CMakeCache.txt` with absolute host paths and a host compiler ABI. It lands at `/ws/src/ros2_pulse/build/`. colcon builds into `/ws/build`, so it is normally inert — but colcon *does* crawl `/ws/src` for packages, and CMake's own `find_package`/`CMAKE_PREFIX_PATH` behaviour around a stale `install/` inside the source tree is exactly the class of thing that produces a build failure whose message points anywhere except the actual cause. When that happens, the harness reports `tail -30 build.log` and `exit 2`, and the reader has to work out that the failure came from their own untracked host directory.

That second point matters here specifically because the script's design goal, stated in its own header, is that "the repo is mounted read-only and copied, so a run can never dirty the working tree". Isolation from the host tree is the intent; copying host build artifacts is the one hole in it.

### Recommendation

Copy only what a build needs. Either the tracked tree (exactly reproducible, but skips uncommitted work — usually wrong for a dev harness):

```bash
git -C /pulse_src archive --format=tar HEAD | tar -x -C /ws/src/ros2_pulse
```

or, keeping uncommitted changes (preferred here — you want to test the working tree):

```bash
mkdir -p /ws/src/ros2_pulse
tar -C /pulse_src --exclude=./.git --exclude=./build --exclude=./install \
    --exclude=./log --exclude=./test/rmw/out -cf - . | tar -xf - -C /ws/src/ros2_pulse
```

`tar` is present in the `ros:<distro>` images, so this needs no extra deb. Worth a one-line comment recording *why* the excludes exist — a stale host `build/` inside the copied source is the failure it prevents.

PR #22, opened 2026-08-01
chore: v0.2.0 release hygiene: CHANGELOG, badges, issue templates, SECURITY.md

TanayK07:
## Summary

Production-readiness item 6. **Do not merge yet, for review.** Suggested order: merge #21 (R1) first, then this, then cut the `v0.2.0` tag (CHANGELOG's 0.2.0 section includes R1).

## What's in

- **CHANGELOG 0.2.0**: both audit rounds (15 issues fixed test-first), jazzy+ publish-side intra, expected-rate alerting + pulse-check, 3-distro CI matrix, honest paired benchmarks replacing the "within noise" claim
- **package.xml**: 0.1.0 → 0.2.0
- **README badges**: live CI status badge; distro badge now `Humble | Jazzy | Kilted`
- **Issue templates** (forms): bug report captures distro/platform/repro/probe-env up front; feature request anchored to the ROADMAP framing; config links private security reporting
- **SECURITY.md**: private advisory path + threat-model notes appropriate for an LD_PRELOAD tool (input surfaces, single write artifact, no sockets, no privileges, CI-pinned 8-symbol surface)

## Not in (deliberate)

- `v0.2.0` tag, cut after both PRs land, so the tag content matches the CHANGELOG
- Demo GIF, needs a recorded terminal session; propose doing it with the launch-assets item (item 8) so the demo shows the final feature set incl. WARN lines

PR #30, opened 2026-08-08
feat: ROS_TOPIC_STATS_FORMAT=jsonl: one JSON object per window (ROADMAP R6)

Tanay10x:
## What

ROADMAP R6: `ROS_TOPIC_STATS_FORMAT=jsonl` emits one JSON object per window (jsonlines.org framing), unblocking the exporter-sidecar story while keeping the probe network-zero and zero-dependency (hand-rolled emitter and a restricted hand-rolled parser, same stance as the YAML-subset spec parser).

Design decisions, each justified at the decision site and tested:

- **`ts_ns` is a decimal string**, epoch ns exceeds 2^53−1, so a JSON number silently corrupts low digits in every double-based consumer (jq, JS). Follows OTLP/JSON's int64-as-string `timeUnixNano` convention; the golden test value is deliberately > 2^53.
- **Absent key = "not measured", never 0**, mirrors `has_pub_max_dt`/`has_recv_max_dt`; a Humble probe never claims rates its tracepoints can't see. Unbounded `max_hz` omits the key (JSON has no Infinity). `topics`/`nodes`/`warns` are always arrays so sidecars need no null guards.
- **Structured warns**, new `evaluateRateSpecWarnings` returns typed violations; the historic string API is now a pure projection over it (pinned byte-identical by the existing R1/R5 tests + an element-for-element equality test), so text and jsonl can never judge differently.
- **RFC 8259 escaping of user-controlled names**, hostile topic names (quote/backslash/newline/control bytes) round-trip; the one-object-per-line framing is asserted independently of the golden bytes.
- **pulse-check learns jsonl**: `parseLog` sniffs a `{` first byte per line (no text line or ROS name can start with `{`). Unknown keys are additive-friendly; records lacking `ts_ns`+`window_s` aren't probe windows, so foreign jsonl exits 2, choosing the exporter format never silently costs CI/watchdog gating.
- **Unknown format value warns once on stderr and falls back to text**; unset/empty stays the silent default with byte-identical output. Banner reports `format=`.
- Integration test validates output with Python's own `json` module, an independent parser proves compliance, not our own reader.

## Verification (TDD)

- Red commit: 18 new tests failing on assertions (goldens, env parse, jsonl reader, structured warns), 105 pre-existing green.
- Green: **123/123** in fast, TSan, and ASan+UBSan lanes; full Humble colcon suite **167 tests, 0 failures** including the new `integration_format`; `pulse-check` on a jsonl log exits 1 with re-derived WARNs, on non-probe jsonl exits 2.


src/core/window_format.cpp
```diff
@@ -113,6 +114,175 @@ auto formatWindow(const std::vector<sTopicStat>& stats, const std::vector<std::s
...
+        appendJsonEscaped(out, s.topic);
+        out += '"';
+        // Key ABSENCE means "not measured", never 0 — the same statement the text format makes
+        // by not printing a line. The pub pair covers both TOPIC and the additive PUB line
+        // (unlike text, intra is included even at 0.0: "no intra publishes observed" is a
+        // genuine count, and a fixed pair keeps the pub schema one shape).
+        if (pub_side) {
+            out += ",\"pub_inter_hz\":";
+            out += sprintfStr("%.6f", s.pub_inter_hz);
+            out += ",\"pub_intra_hz\":";
```
TanayK07, line 199: ## 🔬 GAP ANALYSIS — `pub_intra_hz` is emitted as a measured `0.000000` where the text format emits nothing, breaking the PR's own "absence means not measured" invariant

**Severity:** HIGH — on Humble the intra-publish tracepoint does not exist, so every jsonl window asserts a measured `pub_intra_hz: 0.000000` for a quantity the probe structurally cannot see. This is a wire-format semantic, so fixing it later is a schema break for every sidecar already parsing it.

### The Problem

`formatWindowJsonl` gates the pub *pair* on one boolean:

```cpp
const bool pub_side = TopicRegistry::shouldEmitTopic(s, emit_idle) || s.pub_intra_count > 0;
...
if (pub_side) {
    out += ",\"pub_inter_hz\":";
    out += sprintfStr("%.6f", s.pub_inter_hz);
    out += ",\"pub_intra_hz\":";          // <-- always, even when pub_intra_count == 0
    out += sprintfStr("%.6f", s.pub_intra_hz);
}
```

The text format gates the two values **independently** (`src/core/window_format.cpp:47` and `:57`):

```cpp
if (TopicRegistry::shouldEmitTopic(s, emit_idle)) {   // TOPIC <name> <pub_inter_hz>
if (s.pub_intra_count > 0) {                          // PUB <name> inter=.. intra=..
```

So for the common case — a topic this process published inter-process, `pub_intra_count == 0` — text emits `TOPIC /scan 20.000000` and **no** `PUB` line, while jsonl emits `"pub_intra_hz":0.000000`.

### Proof — pinned by this PR's own golden test

`test/unit/test_window_format_jsonl.cpp:53` builds `/scan` via `stat("/scan", 100, 0, 0, 20.0, 0.0, 0.0)` (`pub_intra_count` never set, i.e. 0) and asserts:

```
{"topic":"/scan","pub_inter_hz":20.000000,"pub_intra_hz":0.000000}
```

The README example in this same diff shows the identical shape. Both the PR description ("a Humble probe never claims rates its tracepoints can't see") and `README.md` ("**An absent key means 'not measured', never 0**: `pub_*`/`recv_*` pairs appear only when the matching text line would") state the opposite rule. The text `PUB` line's own comment in `window_format.cpp:56` says it is *"Emitted only when intra publishes happened this window (impossible on humble: the tracepoint doesn't exist there)"* — which is exactly the case where jsonl now prints a zero.

The `recv` pair does **not** have this problem: `shouldEmitRecv` gates one `RECV inter=.. intra=..` line carrying both values, so the jsonl pair is a faithful mirror. Only the pub pair conflates two independently gated text lines into one gate.

Today nothing crashes: `pulse_check` reads `pub_inter_hz` for `side: pub` rules, so gating is unaffected. What breaks is the consumer contract this format exists to serve:

- On Humble (the distro the README leads with), *every* publishing topic exports `pub_intra_hz = 0` forever. A Prometheus sidecar (the stated R6 consumer) cannot distinguish "no intra-process publishing happened" from "this distro cannot measure intra-process publishing". An alert like `rate(pub_intra_hz) == 0 for 5m` fires on every Humble deployment.
- It is precisely the failure mode `has_pub_max_dt` / `has_recv_max_dt` were introduced to prevent, and this same function honours that rule two lines below (`if (s.has_pub_max_dt)`). The gap fields get presence semantics; the intra rate does not.

The inverse direction is fine and correctly documented: an intra-only publisher (`pub_intra_count > 0`, `pub_inter_count == 0`) emits `"pub_inter_hz":0.000000`, which matches the text `PUB ... inter=0.000000` line byte-for-byte.

### Research

This is the standard distinction between an absent field and a zero-valued one. OpenMetrics/Prometheus makes the same rule normative — an unmeasured series must be *absent*, not exposed as 0, because `0` is a legitimate measured value and downstream `rate()`/`increase()` cannot tell them apart. OTLP applies the same principle via optional fields on `NumberDataPoint`. Your own `has_*_max_dt` design cites this reasoning verbatim in `topic_registry.hpp:63-68`.

### Recommendation

Gate each value on the gate that governs its text twin, keeping the JSON schema additive:

```cpp
const bool pub_inter_side = TopicRegistry::shouldEmitTopic(s, emit_idle);
const bool pub_intra_side = s.pub_intra_count > 0;
const bool pub_side = pub_inter_side || pub_intra_side;
...
if (pub_inter_side) {
    out += ",\"pub_inter_hz\":";
    out += sprintfStr("%.6f", s.pub_inter_hz);
}
if (pub_intra_side) {
    out += ",\"pub_intra_hz\":";
    out += sprintfStr("%.6f", s.pub_intra_hz);
}
```

Note the intra-only case still emits `pub_inter_hz: 0.000000` under this change (because `shouldEmitTopic` returns true only on `pub_inter_count > 0` — so it does *not* fire, and you would need `pub_inter_side = shouldEmitTopic(...) ` plus an explicit decision). Two coherent options:

1. **Presence-strict** (recommended, matches the documented rule): as above, plus update `IntraPublisherCarriesPubPairOnly` to expect `{"topic":"/points","pub_intra_hz":30.000000}` — text's `PUB` line carries `inter=0.000000` only because the line's grammar is fixed; JSON has no such constraint.
2. **Keep the pair, fix the docs**: leave the code and delete the "pairs appear only when the matching text line would" sentence from `README.md` and the PR body, and state plainly that `pub_intra_hz` is `0` on distros without the tracepoint. This is strictly worse for the exporter story but is at least honest.

Either way, add a regression test that asserts a Humble-shaped stat (`pub_inter_count > 0, pub_intra_count == 0`) and a text/jsonl **key-set** equivalence check, so the two encoders' gates can't drift again — the current test suite pins the bytes but never asserts the two formats carry the same set of measurements.

src/core/log_reader.cpp
```diff
@@ -23,6 +26,385 @@ auto statFor(sLogWindow& win, const std::string& topic) -> sTopicStat& {
...
+    const std::from_chars_result res = std::from_chars(c.p, c.end, out);
+    if (res.ec != std::errc{}) {
+        return false;
+    }
+    c.p = res.ptr;
+    return true;
+}
+
+// Skip any well-formed JSON value: how unknown keys stay tolerated without being understood.
+auto skipJsonValue(sCursor& c) -> bool {  // NOLINT(misc-no-recursion) — nesting is user-bounded
```
TanayK07, line 170: ## 🔬 ISSUE — unbounded recursion in `skipJsonValue`: a ~100 KB log line segfaults `pulse-check` (SIGSEGV, exit 139)

**Severity:** HIGH — a malformed/hostile jsonl log crashes the CI/watchdog gate instead of being skipped, defeating both "rejects the line on any malformation" and the project's never-crash stance.

### The Problem

```cpp
auto skipJsonValue(sCursor& c) -> bool {  // NOLINT(misc-no-recursion) — nesting is user-bounded
```

The suppression comment asserts the nesting depth is user-bounded. It is not. `skipJsonValue` recurses once per `{` or `[` with **no depth counter and no limit**, and its input is an arbitrary file handed to `pulse-check` on the command line (`tools/pulse_check.cpp` reads whatever paths the operator or CI passes). Every unknown top-level key routes here (`log_reader.cpp:337` — the `warns` array included, by design), as does every unknown key inside a topic object (`log_reader.cpp:527` in the diff).

`misc-no-recursion` is a clang-tidy *readability* check; silencing it does not add a bound. The only real bound would be an explicit depth counter.

### Proof

Built the PR's `src/core/log_reader.cpp` unmodified against a driver that calls `parseLog` on a single line of the form `{"ts_ns":"1","window_s":1.000,"x":[[[…]]]}`:

```
$ g++ -std=c++17 -O2 -Iinclude src/core/log_reader.cpp drv.cpp -o drv
$ ./drv deep 50000   ; echo exit=$?     # 100 KB line
exit=0
$ ./drv deep 100000  ; echo exit=$?     # 200 KB line
Segmentation fault (core dumped)
exit=139
$ ./drv deep 200000  ; echo exit=$?
Segmentation fault (core dumped)
exit=139
```

At `-O0` (the debug/ASan lanes) the threshold is lower still — depth 200 000 crashes, and the frame is larger, so a smaller file suffices. The crash is a stack overflow: ~8 MB default `RLIMIT_STACK` divided by the frame size of `skipJsonValue`. A ~200 KB input file is well inside "a log someone tailed into the wrong path".

Note this is not caught by the existing tests: `JsonlMalformedSkippedAndUnknownKeysTolerated` covers *shallow* malformation only, and ASan/UBSan do not detect stack exhaustion (only `-fstack-check`/guard pages do, which turn it into the same SIGSEGV).

The probe itself never emits nesting deeper than 2, so a healthy pipeline never hits it. It breaks when:

- an operator points `pulse-check` at a mixed/aggregated log directory containing some other tool's jsonl (the exact scenario `test_pulse_check_gates_jsonl_logs` says must exit 2, not crash);
- a log shipper re-ingests and re-nests records;
- anything untrusted reaches the file — jsonl's whole selling point in this PR is that third parties write and read it.

A `pulse-check` that dies with 139 instead of 0/1/2 turns a health gate into an unclassifiable CI failure, and the "never a silent exit 0 on bad input" guarantee becomes "sometimes exit 139".

### Research

Every hardened JSON parser bounds nesting explicitly, because RFC 8259 §9 says so directly: *"An implementation may set limits on the maximum depth of nesting."* Concretely — RapidJSON exposes `kParseIterativeFlag` for a non-recursive parser, nlohmann/json throws `parse_error.101` past its depth budget, Python's `json` inherits the interpreter recursion limit (`RecursionError`, not a crash), and Go's `encoding/json` caps at 10 000 (golang/go#31789, fixed exactly because deep nesting crashed the runtime). This class of bug has a CWE: **CWE-674 Uncontrolled Recursion**, and it is how CVE-2022-1471-style parser DoS lands in practice.

### Recommendation

Thread a depth budget through and reject past it. Six lines, no behaviour change for real records (the emitter's max depth is 2):

```cpp
// RFC 8259 §9 permits a depth limit; the emitter never nests past 2, so anything
// deeper is either a foreign writer or an attack. Bounded so a hostile line is
// REJECTED (one lost window) instead of overflowing pulse-check's stack.
constexpr int kMaxJsonDepth = 64;

auto skipJsonValue(sCursor& c, int depth = 0) -> bool {  // NOLINT(misc-no-recursion)
    if (depth > kMaxJsonDepth) {
        return false;
    }
    ...
            if (!skipJsonValue(c, depth + 1)) {
                return false;
            }
```

and pass `depth + 1` from the two call sites in `parseTopicObject` / `parseJsonlWindow` (they are already at depth 1 and 2 respectively). Add a regression test alongside the existing malformed-line cases:

```cpp
// Deep nesting is rejected, not fatal: pulse-check must never die on a log file.
TEST(LogReader, JsonlDeepNestingRejectedNotFatal) {
    std::string line = "{\"ts_ns\":\"1\",\"window_s\":1.0,\"x\":";
    line.append(100000, '[');
    line.append(100000, ']');
    line += "}\n";
    EXPECT_TRUE(parseLog(line).empty());
}
```

Also drop the "nesting is user-bounded" justification from the NOLINT comment once the real bound exists — the comment should point at `kMaxJsonDepth`.

src/core/log_reader.cpp
```diff
@@ -23,6 +26,385 @@ auto statFor(sLogWindow& win, const std::string& topic) -> sTopicStat& {
...
+        }
+    }
+    if (!have_name) {
+        return false;  // a rates entry with no topic is not attributable to anything
+    }
+    // Merge via statFor like the text lines do — the emitter writes one entry per topic, but a
+    // foreign/duplicated record should still coalesce rather than shadow.
+    auto& s = statFor(win, tmp.topic);
+    const std::string name = s.topic;
+    s = tmp;
```
TanayK07, line 297: ## 🔬 ISSUE — `parseTopicObject` clobbers instead of coalescing: a repeated topic entry silently zeroes previously-parsed rates and *unsets* `has_*_max_dt`

**Severity:** MEDIUM — the code does the opposite of what its own comment promises, and the failure is silent: rates become 0 (false `WARN`) or a measured gap becomes "unmeasured" (spurious exit 2).

### The Problem

```cpp
// Merge via statFor like the text lines do — the emitter writes one entry per topic, but a
// foreign/duplicated record should still coalesce rather than shadow.
auto& s = statFor(win, tmp.topic);
const std::string name = s.topic;
s = tmp;          // <-- whole-struct overwrite, not a merge
s.topic = name;
```

`tmp` is a **default-constructed** `sTopicStat` into which only the keys present in *this* object were written. `s = tmp` therefore overwrites every field of the existing entry with `tmp`'s defaults for every key this object omitted — including the `has_pub_max_dt` / `has_recv_max_dt` presence flags the surrounding comment block explicitly calls "the measurement marker".

The text parser it claims to mirror does the opposite: `TOPIC`/`PUB`/`RECV`/`JITTER` handlers each assign only their own fields into `statFor(*cur, name)` (`log_reader.cpp:454-473`), which is a genuine field-wise merge.

### Proof

Compiled this PR's `src/core/log_reader.cpp` unmodified and fed it one record with two entries for `/x` — the first carrying pub rate + gap, the second carrying recv rate:

```
{"ts_ns":"1","window_s":1.000,"topics":[
  {"topic":"/x","pub_inter_hz":10.000000,"pub_max_dt_ms":5.000},
  {"topic":"/x","recv_inter_hz":7.000000,"recv_intra_hz":0.000000,"recv_endpoint_seen":true}]}
```

Result:

```
topic=/x pub_inter=0.000000 recv_inter=7.000000 has_pub_dt=0 pub_dt=0.000000
stats=1
```

`pub_inter_hz` went `10.0 -> 0.0` and `has_pub_max_dt` went `true -> false`. Neither is a merge; the second entry shadowed the first, which is exactly what the comment says must not happen.

Today the emitter writes one entry per topic per record, so the path is dead for self-produced logs — which is also why no test covers it. It becomes live under any of these, all of which this PR explicitly designs for:

- **Foreign writers.** The whole point of the jsonl format is that other tools read *and write* it; the parser's stated contract is that a foreign record "should still coalesce".
- **An additive future emitter.** The most natural evolution of this schema — one entry per `(topic, side)` so pub and recv rows can be shipped independently — silently zeroes half of every topic's data when read back by this parser. That is the "fragile pattern" category: correct today, wrong after a change the schema invites.
- **Log concatenation/repair tooling** that merges partial records.

The downstream damage is a *wrong verdict*, not a parse error: a zeroed `pub_inter_hz` makes `evaluateRateSpec` emit `WARN TOPIC /x hz=0.000000` (false alarm, `pulse-check` exit 1), and a cleared `has_pub_max_dt` sends the R5 unmeasured-gap path to exit 2 on a log that did contain the measurement.

### Research

This is the standard "last-write-wins vs. field-mask merge" distinction. Protobuf's JSON mapping and `FieldMask` semantics (google/protobuf `field_mask.proto`) define merge as *only the present fields are applied*, which is precisely the rule your text parser follows and this one does not. JSON Merge Patch (RFC 7386) codifies the same: absent members leave the target unchanged. Prometheus' and OTel's collectors both merge label-identified samples field-wise for the same reason — an absent field must never be able to erase a present one.

### Recommendation

Apply only the fields this object actually carried, mirroring the text handlers. Track presence while parsing and assign selectively:

```cpp
auto& s = statFor(win, tmp.topic);
if (have_pub_inter)  { s.pub_inter_hz = tmp.pub_inter_hz; }
if (have_pub_intra)  { s.pub_intra_hz = tmp.pub_intra_hz; }
if (have_recv_inter) { s.recv_inter_hz = tmp.recv_inter_hz; }
if (have_recv_intra) { s.recv_intra_hz = tmp.recv_intra_hz; }
if (have_recv_seen)  { s.recv_endpoint_seen = tmp.recv_endpoint_seen; }
if (tmp.has_pub_max_dt)  { s.pub_max_dt_ms = tmp.pub_max_dt_ms;  s.has_pub_max_dt = true; }
if (tmp.has_recv_max_dt) { s.recv_max_dt_ms = tmp.recv_max_dt_ms; s.has_recv_max_dt = true; }
```

The two `has_*` flags need no extra bookkeeping — they are already set by their own branches, so they double as their own presence markers (same trick the emitter uses). The four rate flags and `recv_endpoint_seen` need one `bool` each next to `have_name`; alternatively make `tmp` hold `std::optional<double>` for the five and collapse the block.

Then make the promise testable — this is currently the only stated behaviour in the jsonl reader with no test at all:

```cpp
// Duplicated entries for one topic COALESCE field-wise: an absent key must never erase a
// value (or a has_* measurement marker) an earlier entry carried.
TEST(LogReader, JsonlDuplicateTopicEntriesCoalesce) {
    const auto w = parseLog(
        "{\"ts_ns\":\"1\",\"window_s\":1.000,\"topics\":["
        "{\"topic\":\"/x\",\"pub_inter_hz\":10.000000,\"pub_max_dt_ms\":5.000},"
        "{\"topic\":\"/x\",\"recv_inter_hz\":7.000000}]}\n");
    ASSERT_EQ(w.size(), 1u);
    ASSERT_EQ(w[0].stats.size(), 1u);
    EXPECT_DOUBLE_EQ(w[0].stats[0].pub_inter_hz, 10.0);
    EXPECT_DOUBLE_EQ(w[0].stats[0].recv_inter_hz, 7.0);
    EXPECT_TRUE(w[0].stats[0].has_pub_max_dt);
}
```

If you would rather not support coalescing at all, the honest alternative is to *reject* a duplicated topic within one record (`return false`) and delete the comment — but silently shadowing is the one option that matches neither.

src/core/log_reader.cpp
```diff
@@ -23,6 +26,385 @@ auto statFor(sLogWindow& win, const std::string& topic) -> sTopicStat& {
...
+            default:
+                return false;
+        }
+    }
+    return false;  // unterminated
+}
+
+auto parseJsonNumber(sCursor& c, double& out) -> bool {
+    skipWs(c);
+    const std::from_chars_result res = std::from_chars(c.p, c.end, out);
```
TanayK07, line 161: ## 🔬 ISSUE — `parseJsonNumber` accepts `inf`/`nan`, which RFC 8259 forbids: yields undefined behaviour on `ts_ns` and a silent green verdict on NaN rates

**Severity:** MEDIUM — undefined behaviour proven with UBSan, plus a `pulse-check` exit-0 on a log that should never gate green. Both are on the "malformed input" path the parser claims to reject.

### The Problem

```cpp
auto parseJsonNumber(sCursor& c, double& out) -> bool {
    skipWs(c);
    const std::from_chars_result res = std::from_chars(c.p, c.end, out);
    if (res.ec != std::errc{}) {
        return false;
    }
```

`std::from_chars` with the default `chars_format::general` mirrors `strtod` minus hex and leading `+`, and libstdc++ **does** accept `inf`, `infinity` and `nan` (all case-insensitive). RFC 8259 §6 admits no such literals — `Infinity` and `NaN` are exactly the two values the spec calls out as non-representable, which is why this very PR omits an unbounded `max_hz` key rather than writing `inf`. The reader is more permissive than the format it is written for, and this is the function guarding *every* numeric field: `window_s`, all four rates, both gap fields, and the numeric `ts_ns` fallback.

The `ts_ns` fallback then does an unguarded float-to-integer conversion:

```cpp
double d = 0.0;
if (!parseJsonNumber(c, d)) {
    return false;
}
win.ts_ns = static_cast<long long>(d);   // src/core/log_reader.cpp:341
```

`[conv.fpint]/1`: if the truncated value cannot be represented in the destination type, the behaviour is **undefined**. `inf` cannot.

### Proof

Compiled this PR's `src/core/log_reader.cpp` unmodified, with `-fsanitize=float-cast-overflow -fno-sanitize-recover=all`, parsing `{"ts_ns":inf,"window_s":1.0}`:

```
src/core/log_reader.cpp:341:56: runtime error: inf is outside the range of representable
values of type 'long long int'
exit=1
```

Without the sanitizer, the same input parses "successfully" and yields a window with `ts_ns = -9223372036854775808`. Second case, `{"ts_ns":"1","window_s":inf,"topics":[{"topic":"/x","pub_inter_hz":nan}]}`:

```
windows=1
window_s=inf stats=1 hz=nan
```

Note the PR states it runs an ASan+UBSan lane and stayed clean — it does, because no test feeds these tokens. The malformed-input test (`JsonlMalformedSkippedAndUnknownKeysTolerated`) covers structural malformation only, never a malformed *token*.

The probe cannot produce these itself — `TopicRegistry::snapshot` guards the divisor (`const double w = window_s > 0.0 ? window_s : 1.0;`, `topic_registry.cpp:369`), so no emitted rate is ever `nan`. The exposure is entirely on the read side, which is the side this PR opens to third parties.

The consequential failure is the NaN rate, because IEEE-754 comparisons against NaN are all false:

```cpp
if (hz < rule.min_hz || hz > rule.max_hz) {   // rate_spec.cpp:401 — false when hz is NaN
```

So a jsonl log carrying `"pub_inter_hz":nan` produces **no warning at all**, and `pulse-check` exits 0. That is the precise outcome `test_pulse_check_gates_jsonl_logs` is written to forbid ("NEVER a silent exit 0 on bad input") — a corrupt or foreign log gates *green* rather than exit 1 or exit 2. A watchdog wired to this is worse than no watchdog.

### Research

- **RFC 8259 §6:** the grammar for `number` is `[-] int [frac] [exp]`; there is no production for `inf`/`nan`, and §6's interoperability note discusses precisely why implementations must not invent them.
- **libstdc++ `from_chars`:** implemented over the fast_float algorithm, whose `parse_infnan` path accepts `inf`/`infinity`/`nan` with optional sign — matching `strtod` per [charconv.from.chars]/6 ("as if by `strtod` … except that … the plus sign is not recognized and hex is not recognized when `fmt` is `general`"). Infinity/NaN recognition is *not* among the listed exceptions, so acceptance is conforming, not a libstdc++ quirk.
- Hand-rolled parsers in the same position bound this explicitly: RapidJSON gates them behind `kParseNanAndInfFlag` (off by default), serde_json rejects them outright, and Python's `json` accepts only the exact capitalised `NaN`/`Infinity` extensions — never the lowercase forms glibc produces, so a record this reader accepts would be rejected by the very parser your integration test uses as the independent oracle.

### Recommendation

Reject the non-JSON tokens at the one chokepoint, and range-check the integer conversion:

```cpp
auto parseJsonNumber(sCursor& c, double& out) -> bool {
    skipWs(c);
    // RFC 8259 §6 has no inf/nan production, but from_chars mirrors strtod and accepts both.
    // A NaN rate is worse than a rejected line: every comparison in evaluateRateSpec is false,
    // so it would gate GREEN. Reject the token, drop the window (one lost record, like a torn
    // text block).
    if (c.p != c.end && (*c.p == 'i' || *c.p == 'I' || *c.p == 'n' || *c.p == 'N' ||
                         ((*c.p == '-') && c.end - c.p > 1 &&
                          (c.p[1] == 'i' || c.p[1] == 'I')))) {
        return false;
    }
    const std::from_chars_result res = std::from_chars(c.p, c.end, out);
    if (res.ec != std::errc{} || !std::isfinite(out)) {   // belt and braces (1e400 -> inf)
        return false;
    }
    c.p = res.ptr;
    return true;
}
```

`!std::isfinite(out)` alone also closes the overflow route (`1e400` parses to `inf` with `errc{}` on libstdc++ — `from_chars` only reports `result_out_of_range` for values it declines to round). Note `parseNonNegative` in `rate_spec.cpp:56` already applies exactly this `std::isfinite` guard for spec files, with the same rationale — this is the missing half of an existing house rule.

For `ts_ns`, prefer `std::from_chars` into `long long` on the numeric branch too, so an out-of-range value is a clean rejection rather than a cast; if you keep the double path, guard it:

```cpp
if (!std::isfinite(d) || d < -9.2e18 || d > 9.2e18) {
    return false;
}
win.ts_ns = static_cast<long long>(d);
```

Worth adding to `JsonlMalformedSkippedAndUnknownKeysTolerated`:

```cpp
EXPECT_TRUE(parseLog("{\"ts_ns\":inf,\"window_s\":1.0}\n").empty());
EXPECT_TRUE(parseLog("{\"ts_ns\":\"1\",\"window_s\":nan}\n").empty());
EXPECT_TRUE(parseLog("{\"ts_ns\":\"1\",\"window_s\":1.0,\"topics\":"
                     "[{\"topic\":\"/x\",\"pub_inter_hz\":nan}]}\n").empty());
```

src/core/window_format.cpp
```diff
@@ -113,6 +114,175 @@ auto formatWindow(const std::vector<sTopicStat>& stats, const std::vector<std::s
...
+            case '\r': out += "\\r"; break;
+            case '\t': out += "\\t"; break;
+            default:
+                if (static_cast<unsigned char>(c) < 0x20) {
+                    char buf[8];
+                    std::snprintf(buf, sizeof(buf), "\\u%04x",
+                                  static_cast<unsigned>(static_cast<unsigned char>(c)));
+                    out += buf;
+                } else {
+                    out += c;
```
TanayK07, line 141: ## 🔬 GAP ANALYSIS — `appendJsonEscaped` passes non-UTF-8 bytes through verbatim, so a hostile name still produces a record that is not valid JSON *text*

**Severity:** LOW — low reachability (rcl validates node/topic names), but it is the one hole in a claim the README, the PR body and the header comment all make unconditionally, and the failure mode is a whole unreadable file rather than one bad record.

### The Problem

```cpp
default:
    if (static_cast<unsigned char>(c) < 0x20) {
        ... "\\u%04x" ...
    } else {
        out += c;     // every byte >= 0x20 passes through, valid UTF-8 or not
    }
```

The header comment states the intent: *"Everything >= 0x20 passes through verbatim — jsonl is UTF-8 and multi-byte sequences need no escaping."* That is true for **well-formed** UTF-8. It is not true for arbitrary bytes: a lone `0xFF` (or a truncated multi-byte sequence, or a bare surrogate encoded as CESU-8) is `>= 0x20` and is copied straight into the record.

RFC 8259 §8.1 is normative here: *"JSON text exchanged between systems that are not part of a closed ecosystem MUST be encoded using UTF-8."* A record containing `0xFF` is not JSON text at all, regardless of how well the quotes are balanced.

### Proof

Built this PR's `formatWindowJsonl` unmodified with a topic name `"/sc\xFFan"`:

```
$ python3 -c "open('bad.jsonl').read()"
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 54: invalid start byte
$ python3 -c "import json; json.loads(open('bad.jsonl','rb').read().decode('utf-8'))"
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 54: invalid start byte
```

That is the *exact* oracle `test/integration/test_format.py::_jsonl_windows` uses — and it fails at `open()`, before `json.loads` is even reached, so **one** bad byte anywhere in the file makes the whole file unreadable to that consumer, not just its own line. `jq` is more forgiving but lossy: it substitutes U+FFFD, so the name silently changes on the way through (`"/sc�an"`).

For contrast, the framing claim does hold — the line count is unaffected — so this is narrowly about *validity*, not about the `\n` framing the escaping was designed to protect.

Today: essentially unreachable in a healthy stack. `rcl_validate_topic_name` / `rmw_validate_node_name` restrict names to `[A-Za-z0-9_/~{}]`, and this probe reads names from tracepoint payloads that rcl has already validated, so real ROS 2 traffic can never carry a non-ASCII byte here.

It breaks when the name does not come from validated rcl: a caller that bypasses validation, a corrupted tracepoint payload, a future field carrying free-form text (callback names are literally the next R6 item — `rclcpp_callback_register` labels are user-supplied strings, not validated names), or a non-rcl producer of this format. The text format has no equivalent exposure because it is byte-transparent by construction; jsonl is the format that makes encoding a correctness property.

The asymmetry worth noting: this diff *does* harden against the low-probability hostile cases (control bytes, quotes, backslashes, tested in `JsonlHostileTopicNameRoundTrips`) and even ships a reader that decodes surrogate pairs "so a foreign writer's record should still round-trip". Given that level of care on the escape path, the encoding hole reads as an oversight rather than a decision.

### Research

- **RFC 8259 §8.1** (MUST be UTF-8) and **§7** (a string is a sequence of Unicode code points; any code point MAY be escaped as `\uXXXX`).
- **Unicode UTS #22 / the WHATWG Encoding Standard** define the two conforming options for ill-formed input: reject, or substitute U+FFFD. Silent passthrough is neither.
- Serializers in the same position all pick one: Go's `encoding/json` substitutes U+FFFD for invalid bytes (documented in `Marshal`); Rust's `serde_json` cannot hit the case (`String` is UTF-8 by type); RapidJSON's `Writer` has `kWriteValidateEncodingFlag` and fails the write otherwise. None emit the raw byte.

### Recommendation

Escape ill-formed bytes rather than dropping or substituting them — it keeps the record valid *and* keeps `parseLog`'s existing round-trip test passing, because `ÿ` decodes back through `appendUtf8` to... a different byte. So pick deliberately between the two honest options:

**Option A — validate and substitute (recommended; matches Go, one-line-ish, no round-trip claim):**

```cpp
// RFC 8259 §8.1: JSON text MUST be UTF-8. Names reaching here are rcl-validated ASCII in
// practice, but a byte that is not part of a well-formed sequence would make the whole FILE
// undecodable for a UTF-8 consumer (python's json fails at open(), not at the bad line).
// Substitute U+FFFD, the Unicode-conforming choice, so one bad name costs one name.
```
validate each multi-byte sequence with a small scanner and emit `\xEF\xBF\xBD` for any ill-formed byte. Then relax `JsonlHostileTopicNameRoundTrips` to assert byte-identical round-trip for *well-formed* names only, and add a case asserting an ill-formed name yields a decodable record.

**Option B — keep byte transparency, drop the claim:** leave the code and amend `README.md` ("names are escaped per RFC 8259, so a hostile name can never break the one-object-per-line framing") to say *framing* explicitly and note that names are emitted as raw bytes, so a producer that does not validate names can emit non-UTF-8 records. Cheaper, but it makes the format's validity conditional on something outside the format.

Either way, the header comment in `window_format.cpp:120-123` should stop asserting "jsonl is UTF-8" as a property of the output when the output only inherits it from the input.

PR #32, opened 2026-08-20
feat(pulse-top): live Textual TUI over the probe's jsonl log

Tanay10x:
Implements the Option A (Textual) direction picked from the design canvas: a live terminal dashboard consuming the v0.3.0 jsonl log. Core functionality untouched, this is a pure file consumer with no ROS dependency and no graph presence (the differentiator vs ros2top/ornis/etc., which join the graph to measure it).

- Topics table: pub/intra/recv Hz, max gap, 60-window sparklines, warn-colored rows; `s` sort cycle, `w` warns-only, row-select detail sidebar
- Tree view (topic namespaces, live rates), Nodes liveness, structured Warns view
- Absence contract preserved end to end: absent jsonl key renders as `-`, never 0
- Reader survives partial writes, truncation/rotation, missing file
- `pulse-top --demo`: scripted-incident generator for the launch GIF
- 20 pytest units + headless Textual pilot smoke (all green locally); new `pulse-top (pytest)` CI lane
- Packaged `ros2-pulse-top`, console script `pulse-top`, `pip install ./tools/pulse-top`


tools/pulse-top/pulse_top/reader.py
```diff
@@ -0,0 +1,41 @@
...
+            size = os.path.getsize(self.path)
+        except OSError:
+            return []
+        if size < self._offset:  # truncated/rotated: start over
+            self._offset = 0
+            self._partial = ""
+        if size == self._offset:
+            return []
+        try:
+            with open(self.path, "r", errors="replace") as f:
```
Tanay10x, line 33: ## ISSUE: text-mode tell() compared against byte size — offset desync on multi-byte UTF-8

**Severity:** HIGH — silent line loss or spurious re-reads when a poll lands mid-UTF-8-sequence

**The Problem**
`poll()` mixes two units: `os.path.getsize()` returns bytes, but `f.tell()` after a text-mode read returns CPython's opaque text-cookie. The truncation check (`size < self._offset`) and the no-change fast path (`size == self._offset`) compare them as if they were the same unit. Worse, `errors="replace"` means a poll that lands in the middle of a multi-byte UTF-8 character decodes the partial bytes to U+FFFD and advances past them — the tail of that character is then decoded as garbage on the next poll.

**Theory / Proof**
The probe's jsonl escapes only quote/backslash/control characters per RFC 8259 (README, "JSON Lines output") — non-ASCII topic and node names pass through as raw multi-byte UTF-8. The probe writes each window with a single write(), but this poller reads on its own clock and can observe a partially flushed page on a different filesystem (NFS-mounted robot logs) or after a crash. CPython's `TextIOWrapper.tell()` happens to equal the byte offset only when no decoder state is pending — exactly the case a mid-character read violates.

**What happens today vs. what will break**
ASCII-only logs on local disk: works, cookies coincide with byte offsets. A topic named with any non-ASCII character (legal in the file, since the probe does not forbid it) plus an unluckily timed poll: one corrupted record (dropped by the tolerant parser — a silently missing window) and a possible offset desync that misclassifies the next poll as truncation, replaying the whole file.

**Research**
`tail -f` implementations (GNU coreutils, pygtail, watchdog-based followers) all track byte offsets on a binary stream and decode after line-splitting, precisely to keep offsets exact and to hold incomplete trailing bytes with the partial line.

**Recommendation**
Track bytes; decode per completed line:
```python
with open(self.path, "rb") as f:
    f.seek(self._offset)
    chunk = f.read()
    self._offset = f.tell()
data = self._partial + chunk
*lines, self._partial = data.split(b"\n")
return [ln.decode("utf-8", errors="replace") for ln in lines if ln]
```
with `self._partial: bytes = b""`. Same tests pass; the partial-line buffer now also holds partial characters.

tools/pulse-top/pulse_top/reader.py
```diff
@@ -0,0 +1,41 @@
...
+                f.seek(self._offset)
+                chunk = f.read()
```
Tanay10x, line 35: ## ISSUE: unbounded catch-up read — UI freeze proportional to log size on attach

**Severity:** MEDIUM — first poll on a long-running probe log blocks the render thread for the whole backlog

**The Problem**
`f.read()` reads everything from the stored offset to EOF in one call, and `poll()` runs inside the Textual `set_interval` tick on the event loop. Attaching `pulse-top` to a probe that has been running for days (a 5 s window, 24 h = ~17k records; with many topics easily tens of MB) parses the entire history inside a single tick before the first frame that reflects it.

**Theory / Proof**
Textual timers run on the asyncio event loop; a synchronous read+parse of N MB blocks all input handling and rendering for its duration. The dashboard only renders the rolling state (60-window deques), so replaying the full history serves nothing beyond warming sparklines.

**What happens today vs. what will break**
`--demo` and fresh logs: instant. Real deployment attach to an old log — the actual operator workflow this tool exists for — multi-second freeze with a blank screen, reading as a hang.

**Research**
`tail -f` semantics: start at EOF (or `-n N` lines back). btop/htop-family tools never replay history on attach; they sample from now.

**Recommendation**
On first attach, seek to the tail minus a bounded window, then stream normally:
```python
TAIL_BYTES = 256 * 1024  # ~60 windows of a large graph
if self._offset == 0 and size > TAIL_BYTES:
    self._offset = size - TAIL_BYTES
    self._skip_first_partial = True   # drop the first (mid-record) line
```
and cap each poll's read (e.g. 1 MB), carrying the rest to the next tick so the UI stays live while catching up.

tools/pulse-top/pulse_top/app.py
```diff
@@ -0,0 +1,275 @@
...
+                for col, cell in zip(COLUMNS, self._row_cells(name, st)):
+                    table.update_cell(name, col, cell)
+        if self._selected is None and want:
+            self._selected = want[0]
+
+    def _refresh_warns_strip(self) -> None:
+        lines = [f"[{DIM} bold]WARNS · structured from jsonl[/]"]
+        for w in self._state.warns:
+            c = WARN if w.kind == "topic_rate" else BAD
+            lines.append(f"[bold {c}]{w.kind}[/] {w.detail}")
```
Tanay10x, line 158: ## ISSUE: log-controlled strings interpolated into Rich markup — MarkupError DoS from a hostile or corrupt log

**Severity:** MEDIUM — a topic/node name containing "[" crashes or restyles the dashboard

**The Problem**
`_refresh_warns_strip`, `_refresh_sidebar`, `_refresh_nodes`, `_refresh_warns_tab` and the tree labels build display strings with f-strings that embed log-derived values (`w.detail`, `self._selected`, node names, namespace parts) directly into Rich markup passed to `Static.update()` / `Text.from_markup()`. Rich parses `[` as a tag opener: a name like `/x[bold red on white]` injects styling, and an unbalanced `[/` raises `MarkupError` inside the render tick. The `DataTable` path is already safe — it wraps values in `Text()` — so the codebase is inconsistent about the same trust boundary.

**Theory / Proof**
ROS 2 topic names are restricted to `[A-Za-z0-9_/~{}]`, so a live well-behaved graph cannot trigger this. But pulse-top's input is a FILE, and the probe's own threat model (README: names "come from USER code: escaped per RFC 8259 so a hostile name can never break the framing") explicitly treats names as attacker-adjacent. The jsonl framing survives a hostile name; the TUI then un-survives it at the render layer. Rich documents `Text()` assembly or `rich.markup.escape()` as the required handling for untrusted text.

**What happens today vs. what will break**
Well-formed robot logs: fine. A corrupt log, a non-ROS producer of the format, or a deliberately crafted file shared for post-mortem: unhandled `MarkupError` inside `set_interval` — the exact "viewer dies on the artifact it exists to inspect" failure the tolerant parser was built to avoid.

**Research**
Textual/Rich guidance is consistent: never feed untrusted text through the markup parser; build `Text` objects (`Text.assemble(("kind", "bold red"), " ", detail)`) or escape first.

**Recommendation**
Route every log-derived string through `rich.markup.escape()` at the single point it enters a markup f-string, or switch these four renderers to `Text.assemble()` like the table already does:
```python
from rich.markup import escape
lines.append(f"[bold {c}]{escape(w.kind)}[/] {escape(w.detail)}")
```

tools/pulse-top/pulse_top/app.py
```diff
@@ -0,0 +1,275 @@
...
+        return (
+            Text(name, style=f"bold {color}" if color else ""),
+            Text(fmt(t.pub_inter_hz), justify="right"),
+            Text(fmt(t.pub_intra_hz), justify="right", style=GOOD if t.pub_intra_hz else DIM),
+            Text(fmt(t.recv_inter_hz), justify="right"),
+            Text(fmt(gap), justify="right", style=f"bold {BAD}" if st.warn_kind == "topic_gap" else DIM),
+            Text(sparkline(st.rate_history), style=color or DIM),
+        )
+
+    def _refresh_table(self) -> None:
```
Tanay10x, line 138: ## GAP ANALYSIS: a stalled topic keeps rendering its last rate as if current — the absence-vs-zero ethos stops at the display layer

**Severity:** MEDIUM — the dashboard shows a dead topic as healthy, which is the one lie this project's docs promise not to tell

**The Problem**
`_refresh_table` renders `st.latest` for every known topic, but `StatsState.apply` only updates `latest` for topics present in the incoming window. A topic that stops publishing (and is not covered by a `topic_rate`/`topic_gap` rule) simply disappears from subsequent windows: its row freezes at the last measured Hz, sparkline flat, no visual difference from a live steady topic.

**Theory / Proof**
The probe deliberately omits silent topics per window (README, `ROS_PULSE_EMIT_IDLE`): absence means "no traffic seen", a strictly stronger statement than any number. The parser and model preserve that (Optionals, em-dash rendering) — then the table erases it by presenting the previous window's measurement without a staleness marker. This is exactly the failure mode the R5 CHANGELOG entry calls out for rate means ("a windowed mean cannot see a stall"), reproduced in the viewer.

**What happens today vs. what will break**
Continuously publishing topics: correct. The interesting case — the operator staring at the dashboard during an incident where a topic dies outright — shows a confident stale number until they notice the timestamp never changes (which is not displayed).

**Research**
htop greys processes that exited between samples; k9s marks resources with an AGE column; Grafana renders "no data" distinctly from the last value precisely because last-value-carried-forward hides outages.

**Recommendation**
Track `last_seen_window` on `TopicState` (set in `apply`); in `_row_cells`, when `state.windows_seen - last_seen_window > 0`, dim the whole row and replace rates with the em dash plus an age tag:
```python
Text(f"— stale {age}w", style=DIM)
```
keeping the sparkline history visible so the moment of death is readable.

tools/pulse-top/pulse_top/app.py
```diff
@@ -0,0 +1,275 @@
...
+            f"[{DIM}]recv inter[/] {fmt(t.recv_inter_hz)} Hz\n"
+            f"[{DIM}]recv intra[/] {fmt(t.recv_intra_hz)} Hz\n"
+            f"[{DIM}]pub gap[/]    {fmt(t.pub_max_dt_ms)} ms\n"
+            f"[{DIM}]recv gap[/]   {fmt(t.recv_max_dt_ms)} ms\n"
+            f"[{DIM}]endpoint[/]   {'seen' if t.recv_endpoint_seen else '—'}"
+        )
+
+    def _refresh_tree(self) -> None:
+        tree = self.query_one("#ns-tree", Tree)
+        tree.clear()
```
Tanay10x, line 187: ## OPTIMIZATION: full Tree teardown/rebuild on every changed tick — selection and expand state destroyed while the user is navigating

**Severity:** LOW — UX degradation on the Tree tab under live data

**The Problem**
`_refresh_tree` runs inside `_refresh_all`, i.e. on every poll that delivered at least one window (every ~1-5 s live). It calls `tree.clear()` and rebuilds every node, which discards the user's cursor position, any collapsed branches, and scroll offset — mid-navigation, once per window.

**Theory / Proof**
Textual's `Tree.clear()` drops all `TreeNode` objects; there is no identity to restore selection against afterwards. The topic SET changes rarely (topology events), while the rates on existing leaves change every window — the rebuild conflates the two cadences.

**What happens today vs. what will break**
Static demo screenshotting: invisible. A live 5 s window: the tree visibly resets under the cursor five times a minute, making the tab unusable for actual browsing on a busy graph.

**Research**
The standard TUI pattern (k9s resource trees, lazygit file trees) is label-refresh-in-place: mutate existing node labels, add/remove only the delta.

**Recommendation**
Cache `self._tree_paths: dict[str, TreeNode]`. Each refresh: if the sorted topic set is unchanged, only `node.set_label(...)` the leaves; otherwise diff — add new paths, `node.remove()` vanished ones. Rebuild-from-scratch only on first population.

tools/pulse-top/pulse_top/model.py
```diff
@@ -0,0 +1,175 @@
...
+        self.window_s = 0.0
+        self.last_ts_ns = 0
+
+    def apply(self, window: Window | None) -> None:
+        if window is None:
+            return
+        self.windows_seen += 1
+        self.window_s = window.window_s
+        self.last_ts_ns = window.ts_ns
+        self.warns = window.warns
```
Tanay10x, line 136: ## ISSUE: warns reflect only the latest window — a one-window transient is a sub-second blink the operator cannot catch

**Severity:** MEDIUM — the highest-value events (brief stalls) are the least visible ones

**The Problem**
`self.warns = window.warns` replaces the active warn set wholesale each window. A `topic_gap` fired for exactly one window (a single 800 ms stall inside one 5 s window — the R5 headline case) is displayed for one window period and vanishes from the strip, the Warns tab, and the row coloring.

**Theory / Proof**
The probe intentionally emits warns per-violating-window (stateless flush-time checks). A monitor layered on that stream needs its own retention to be useful to a human: reaction time to a glanceable dashboard is measured in seconds-to-minutes, not one window. Notably the design mockup this PR implements already renders the retained form ("stall 05:12:41 — gap fired 7 windows ago") — the model never grew the state to back it.

**What happens today vs. what will break**
Sustained violations (demo loop): fine, re-emitted every window. The transient single-window stall — the exact scenario R5 was built to expose — appears and disappears before a human can read it.

**Research**
Alertmanager keeps resolved alerts visible with a "resolved" state and age; btop retains process death markers for several refresh cycles for the same reason.

**Recommendation**
Keep a bounded recent-warns deque alongside the live set:
```python
self.recent_warns: deque = deque(maxlen=50)   # (windows_seen, warn)
for w in window.warns:
    self.recent_warns.append((self.windows_seen, w))
```
Warns tab renders `recent_warns` with an age column ("Nw ago", dimmed when no longer active); the strip and row coloring keep using the live set.

tools/pulse-top/pulse_top/model.py
```diff
@@ -0,0 +1,175 @@
...
+            if st is None:
+                st = TopicState(latest=tw)
+                st.rate_history = deque(maxlen=self._history)
+                self.topics[tw.topic] = st
+            st.latest = tw
+            st.warn_kind = warned.get(tw.topic)
+            st.rate_history.append(st.rate)
+        # A topic absent from this window but warned about (e.g. gap on a stalled
+        # topic that emitted nothing) still carries its warn.
+        for topic, kind in warned.items():
```
Tanay10x, line 150: ## ISSUE: warned-but-absent reconciliation relies on dataclass equality of float-bearing records — fragile and partly dead code

**Severity:** LOW — works today; quietly stops working on reasonable future change

**The Problem**
```python
if topic in self.topics and self.topics[topic].latest not in window.topics:
```
membership on a `list[TopicWindow]` uses dataclass `__eq__` over seven mostly-float fields to answer a question about NAMES ("was this topic in this window?"). For topics present in the window the check is also redundant — the first loop already set `warn_kind` via `warned.get()`. The branch only matters for warned-but-absent topics, and it reaches that answer by comparing float payloads instead of comparing names.

**Theory / Proof**
Two hazards: (1) add any non-comparable or intentionally excluded field to `TopicWindow` (e.g. a parsed-at timestamp, `field(compare=False)` today, forgotten tomorrow) and the equality stops meaning "same record", silently flipping which topics keep their warn; (2) O(topics^2) per window once graphs get large, in the hot apply path.

**What happens today vs. what will break**
Current field set: correct results. First schema evolution of `TopicWindow`: the reconciliation changes behavior with no test failing, because the tests exercise names, not record equality.

**Research**
The C++ side answers the same question with name sets (log_reader merges by topic key); mirroring that keeps the two readers' semantics aligned.

**Recommendation**
Compute the name set once and use it in both loops:
```python
present = {t.topic for t in window.topics}
for topic, kind in warned.items():
    if topic in self.topics and topic not in present:
        self.topics[topic].warn_kind = kind
for name, st in self.topics.items():
    if name not in present and name not in warned:
        st.warn_kind = None
```

tools/pulse-top/pulse_top/reader.py
Tanay10x: Fixed in #33 — reader is binary-mode now, byte offsets throughout, decode per completed line. Added a regression that splits a two-byte UTF-8 char across polls; it held the partial bytes correctly.

tools/pulse-top/pulse_top/reader.py
Tanay10x: Fixed in #33 — first attach on a backlog over 256 KB seeks to tail and drops the mid-record line; every poll reads at most 1 MB and carries the rest to the next tick. Tests cover both.

tools/pulse-top/pulse_top/app.py
Tanay10x: Fixed in #33, and the recommendation here was actually insufficient — rich.markup.escape() still crashed under Textual v6, which parses its own markup dialect (caught by a hostile-name smoke test). Went with Text-object assembly on all four surfaces instead; that's parser-proof regardless of version.

tools/pulse-top/pulse_top/app.py
Tanay10x: Fixed in #33 — TopicState tracks last_seen; rows absent from the current window render dimmed with em-dash rates and a 'stale Nw' tag, sparkline history kept so the moment of death stays readable.

tools/pulse-top/pulse_top/app.py
Tanay10x: Fixed in #33 — tree relabels leaves in place when the topic set is unchanged, full rebuild only on topology change. Cursor and expand state survive live windows now.

tools/pulse-top/pulse_top/model.py
Tanay10x: Fixed in #33 — recent_warns deque (maxlen 50) with fire-window; Warns tab shows live warns bold and history dimmed with 'Nw ago'. The strip and row coloring stay on the live set, so honesty about current state is unchanged.

tools/pulse-top/pulse_top/model.py
Tanay10x: Fixed in #33 — presence and reconciliation both use a name set now, matching how the C++ log_reader keys by topic. Behavior identical today, immune to TopicWindow schema changes.

